PR #9, opened 2015-05-08
Ac dm funct tests

andrewchong:


robyoung:
Could any of these commits be squashed?

features/step_definitions/journey_test_steps.rb
```diff
@@ -0,0 +1,339 @@
+require "ostruct"
+require "rest_client"
+
+store = OpenStruct.new
```
TheDoubleK: Alexey did some work to remove the need for a global store and to use instance variables instead.  It would be good to do the same thing here.

features/step_definitions/journey_test_steps.rb
```diff
@@ -0,0 +1,340 @@
...
+  current_url.should have_content(store.serviceID)
+  page.should have_content('Service attributes')
+  page.should have_content('Description')
+  page.should have_content('Features and benefits')
+  page.should have_content('Pricing')
+  page.should have_content('Documents')
+
+  servicename = find(
+    :xpath,
+    "//table[2]/tbody/tr[1]/td[2]"
```
quis: This drilling down through elements seems quite fragile. Not sure if that's good or bad, maybe something we should talk about as a group?

PR #15, opened 2015-06-08
Changes to implementation of supplier journey; function service_listed_and_in_correct_order

andrewchong:


features/admin_journey.feature
```diff
@@ -13,18 +13,18 @@ Scenario: As an admin user I wish be able to log in and to log out of Digital Ma
 @logout
 Scenario: As an admin user who has logged in to Digital Marketplace, I wish to search for a service
```
TheDoubleK: Should some (most?) of the step definitions be tagged as `staging-only`, as we don't want any of the test services being created on production.

TheDoubleK: If you add the tag then I'll fix the Jenkins job to skip those tests when running against production.

features/admin_journey.feature
```diff
@@ -1,3 +1,4 @@
+@not-production
```
TheDoubleK: Is it possible to tag individual scenarios, or only Features?  It would be nice for the Admin to at least log in and log out to check that the production admin app is up.

features/supplier_journey.feature
```diff
@@ -0,0 +1,23 @@
+@not-production
```
TheDoubleK: And the same here - it would be nice for the supplier to log in and out, even if they don't have any services to edit. (They could maybe change their contact info once that feature is live?)

PR #19, opened 2015-06-17
New admin scenarios for service status changes and renamed/added tags

andrewchong:


features/step_definitions/journey_test_steps.rb
```diff
@@ -2,7 +2,16 @@
...
+  if page.has_content?('Log out') or page.has_content?('Logout')
+    find(
+      :xpath,
+      "//a[contains(@href, '/logout')]"# and contains (text(), 'Log out')]"
+    ).click
+  end
+end
+
+After('@logoutafter') do
   page.click_link_or_button('Log out')
```
allait: Should this be the same as `@logoutbefore`?

PR #43, opened 2015-07-29
add delay in attempt to fix intermittent preview functional test failure

andrewchong:
Getting intermittent failures when page expected is not presented in time. Added "Sleep 1" to delay checking of the page existence


allait:
Is there an example of a failing job I could look at?
Field edits should be synchronous – since we only have one DB server there should be no replication lag and I'd expect the test to read back the change on the next request without any delays.

PR #44, opened 2015-07-31
new admin scenarios added and implemented, some refactoring and some house keeping

andrewchong:
Still to be added: user deactivation and invitation
Added new pdf file for upload in edit documents scenario.
Accounted for change in the admin pages.
NOTE: Second commit with same title was just a change to a scenario name.


TheDoubleK:
Looks good!

PR #45, opened 2015-08-05
new scenarios for admin journey, some refactoring and tidying up

andrewchong:
New environment variables:(Needs to be added to jenkins preview functional test job.
def dm_supplier2_uname()
  ENV['DM_SUPPLIER2EMAIL']
end
def dm_supplier3_uname()
  ENV['DM_SUPPLIER3EMAIL']
end

New fixtures file for test-supplier-user.son

### DIFF
--- a/features/step_definitions/journey_test_steps.rb
+++ b/features/step_definitions/journey_test_steps.rb
@@ -5,7 +5,7 @@ require "rest_client"
 Given /I am on the '(.*)' login page$/ do |user_type|
   if user_type == 'Administrator'
     visit("#{dm_frontend_domain}/admin/login")
-  else user_type == 'Supplier'
+  elsif user_type == 'Supplier'
     visit("#{dm_frontend_domain}/#{user_type.downcase}s/login")
   end
   page.should have_content("#{user_type}" ' login')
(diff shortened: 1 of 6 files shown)


features/step_definitions/journey_test_steps.rb
```diff
@@ -47,6 +68,10 @@
   click_link_or_button(button_name)
 end
 
+When /I click the '(.*)' button for the supplier user '(.*)'$/ do |button_name,user_name|
+  find(:xpath, "//*/span[contains(text(),'#{user_name}')]/../../td/*//button[contains(text(),'#{button_name}')]").click
```
TheDoubleK: This works OK, but is pretty convoluted XPATH.  I can't see a better way to do things with the way the app is at the moment, but maybe we could make a change to the admin app itself to make the right button for a user more easily findable.

For example, we might be able to add a 'name' attribute to each button (e.g. name='Activate user foo@bar.com'), which would mean the tests can just look for that button directly rather than finding the row with the username and navigating from there through the DOM.

@quis - what do you think?  Am I suggesting something really awful?  I appreciate it's not good to add attributes and bloat the code solely to make tests work, but a descriptive `name` (or something) on a button might not be the end of the world, and would make this considerably less brittle?

pcraig3: Yeah, I'm going to have to do something similar.   
Basically, I need to check a user that's been removed by a supplier has an 'activate' (not a 'deactivate' button) in its row.

tombye: I'd be wary of adding things to the HTML which are for the tests. The problem with that is you're trying to make the tests more readable by adding stuff to the HTML that only makes sense if you're trying to read the tests. If I come to the HTML trying to understand something else about it, I'll be wondering why those things are there because they only make sense from the perspective of the tests.

In my opinion this is a limitation of our testing language only using XPath. Could we not break it up a bit more into helper functions so it better describes the bit you're testing as you would describe it yourself?

For example, I'd say it as: ' the activate button for the email row of the supplier table' which we could split into these helpers:

1. get supplier table (get table with caption of 'supplier')
2. get table row for email field (get row of table where 1st cell contains 'email' text)
3. get activate button for a table row (get button in last cell of table row)

get_table_called('supplier').get_row_for('email').get_button('activate')

?

tombye:
While the points about the bloated XPath are valid, the discussion around this is holding up this being merged when the code works and all the tests pass.

I was part of that discussion so, to be clear, I agree we should look at making the XPath more readable and less brittle but that can be dealt with separately.

PR #46, opened 2015-08-06
Make xpath for supplier update more specific

pcraig3:
The xpath was too general and matching multiple elements on the supplier dashboard page.
This ties the xpath lookup to the table specified by the `service_path`.

Fixes this bug:
screen shot 2015-08-06 at 14 24 16


tombye:
:+1:

PR #48, opened 2015-08-06
[##97631496] Add SSP create-draft features for all lots

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/97631496

This adds features for all lots that create a new draft, complete all the pages (with a "happy path" - should be no errors), navigate to the summary page and delete the service to clean up.

The IaaS feature additionally checks that the correct values are displayed on the summary page - this is the same as we had for the G6 tests.  We might want to add summary page checking to the other lots too, but it would probably be overkill.

I have also copied the validations feature from the G6 tests, but have not had time to fix it to work with the new page and field names, hence the whole feature is marked as WIP.  I have only included the `ssp_steps` needed for the happy-path journeys here, so implementing the validation checking will probably require extra steps that can be found in the deleted methods here: https://github.com/alphagov/digitalmarketplace-smokey/commit/d12c2153d3f76507d1ed8a76487de7d785f4b6d1)

### DIFF
--- a/features/ssp_iaas_service.feature
+++ b/features/ssp_iaas_service.feature
@@ -1,0 +1,489 @@
+@not-production @functional-test
+Feature: Submitting a new service for IaaS
+  In order to submit my services as a supplier user
+  I want to answer questions about my service
+
+  Background: Log in as a supplier
+    Given I am on the 'Supplier' login page
+    When I login as a 'Supplier' user
+    Then I should be on the supplier home page
+
+  Scenario: Select lot
+    Given I am at '/suppliers'
+    When I click 'Apply to become a G-Cloud 7 supplier and add services'
+    And I click 'Add a service'
+    When I choose 'Infrastructure as a Service (IaaS)'
+    And I click 'Save and continue'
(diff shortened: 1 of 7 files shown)


features/ssp_iaas_service.feature
```diff
@@ -0,0 +1,489 @@
...
+
+  Background: Log in as a supplier
+    Given I am on the 'Supplier' login page
+    When I login as a 'Supplier' user
+    Then I should be on the supplier home page
+
+  Scenario: Select lot
+    Given I am at '/suppliers'
+    When I click 'Apply to become a G-Cloud 7 supplier and add services'
+    And I click 'Add a service'
```
quis: This is going to fail because the 'Add service' link is no longer on this page.

TheDoubleK: Well it was there yesterday afternoon!

quis: Haha, sorry Kev. Consider it a note to self to fix this.

quis: Should be addressed by 409fdee

PR #49, opened 2015-08-10
Fix smoke tests race condition on summary text check

allait:
Capybara current_url access doesn't wait for page to load, so the
URL comparison can fail due to page load timing issues.

Using `find` blocks the step until the element is present on a page,
at which point the page was loaded and the `current_url` should be
up to date.

### DIFF
--- a/features/step_definitions/journey_test_steps.rb
+++ b/features/step_definitions/journey_test_steps.rb
@@ -877,10 +877,10 @@ Then /I am on a page with that service\.(.*) in search summary text$/ do |attr_n
 end
 
 Then /I am on a page with '(.*)' in search summary text$/ do |value|
+  find(:xpath, "//*[@class='search-summary']/em[1]").text().should == normalize_whitespace(value)
+
   query_string = CGI.escape value
   current_url.should include("q=#{query_string}")
-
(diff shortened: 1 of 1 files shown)


features/step_definitions/journey_test_steps.rb
```diff
@@ -877,10 +877,10 @@ def filter_to_check(filter_name,filter_value,filter_exist)
 end
 
 Then /I am on a page with '(.*)' in search summary text$/ do |value|
+  find(:xpath, "//*[@class='search-summary']/em[1]").text().should == normalize_whitespace(value)
```
quis: Might be a good idea to leave a comment here because the implication of the change is non-obvious?

allait: There's an explanation in the commit message. It's not the first time we run into this, but I don't think a comment with "find should be the first check" in this step would help with similar issues in others (i.e. it won't be any more noticeable than a commit message)

quis: I meant to prevent anyone making changes later inadvertently reintroducing an intermittent failure. But I'm willing to accept that a commit message does that.

quis:
:+1:

PR #50, opened 2015-08-10
New scenario-view of contributors on admin deactivating/reactivating a user

andrewchong:
Added - fix for intermittent failure issue on preview.


features/step_definitions/journey_test_steps.rb
```diff
@@ -5,10 +5,11 @@
 Given /I am on the '(.*)' login page$/ do |user_type|
   if user_type == 'Administrator'
     visit("#{dm_frontend_domain}/admin/login")
+    page.should have_content("#{user_type}" ' login')
```
robyoung: I think this would be clearer as `"#{user_type} login"`

robyoung:
:+1:

PR #52, opened 2015-08-11
Fix radio set choose tests

robyoung:
Get rid of the "I choose ''" step in favour of the "I choose '' for ''" step. Having both was causing errors because it was ambiguous which one to use and the later is more specific.


andrewchong:
There are changes required for ssp_validation.feature as well.

robyoung:
Having looked at that a bit closer, the validation feature has not been converted to the new SSP yet.

PR #53, opened 2015-08-11
New scenarios: viewing and deactivating contributors

pcraig3:
Added one scenario whereby a logged-in supplier user sees their other contributors, and another where this selfsame user gets rid of one.  
In the second scenario, we also make sure that the inactive user can't log in and that the admin sees them as being inactive.

### DIFF
--- a/features/admin_journey.feature
+++ b/features/admin_journey.feature
@@ -11,7 +11,7 @@ Scenario: Setup for tests
 Scenario: As an admin user I wish be able to log in and to log out of Digital Marketplace
   Given I am on the 'Administrator' login page
   When I login as a 'Administrator' user
-  Then I am presented with the 'Admin' page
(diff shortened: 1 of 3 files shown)


features/admin_journey.feature
```diff
@@ -11,7 +11,7 @@ Scenario: Setup for tests
 Scenario: As an admin user I wish be able to log in and to log out of Digital Marketplace
   Given I am on the 'Administrator' login page
   When I login as a 'Administrator' user
-  Then I am presented with the 'Admin' page
+  Then I am presented with the admin 'Admin' page
```
robyoung: Then I am presented with the admin admin page?

robyoung:
Looks good to me. Would be good to get a review from @andrewchong

andrewchong:
:thumbsup:

PR #55, opened 2015-08-14
Fix SSP service description tests

robyoung:
The service description has been split over two pages. One for service name and one for the description.


andrewchong:
:thumbsup:

PR #69, opened 2015-09-09
Fix unsuccessful login message

pcraig3:
After changing the login message, some of the functional tests started breaking.

So I fixed it.


pcraig3:
:+1:

PR #71, opened 2015-09-14
Standardise environment variables

robyoung:
This commit does two things but they're a bit tricky to pick apart.

1. Remove use of `eval`

`eval` is considered harmful because it allows arbitrary execution of code from variables that would not necessarily be expected to be running code. Most of the uses were just calling the functions they represent however for the dynamic domain and access token uses two new functions had to be introduced to resolve the domain and access tokens for the specific apps.

2. Rename environment variables to closer match apps

Rename the environment variables to closer match the field names used in the frontend applications. This makes it clearer what the environment variables represent when adding them to the Jenkins configuration. Also, making the function names match the environment variables they wrap
makes it easier to follow where these variables are being used.


quis:
:+1:

features/support/env.rb
```diff
@@ -40,28 +66,28 @@ def dm_frontend_domain()
   ENV['DM_FRONTEND_DOMAIN']
 end
 
-def dm_admin_uname()
-  ENV['DM_ADMINISTRATORNAME']
+def dm_admin_email()
+  ENV['DM_ADMINISTRATORNAME'] || ENV['DM_ADMIN_EMAIL']
```
TheDoubleK: Are these alternates for old Jenkins vs. new Jenkins?  I'd be tempted to just have the new ones and fix the old Jenkins jobs, but I guess either way is fine (i.e. we can remove the alternates once old Jenkins is turned off) - as long as we remember to tidy up here once old Jenkins is off.

robyoung: The reason why I kept both was so that we wouldn't have failing checks between this getting merged and the jobs being changed. I won't be letting the old environment variables hang around for long.

PR #77, opened 2015-09-21
Make smoke tests use functional test user

robyoung:
In migrating to the new jenkins machine the users used by the smoke tests and functional tests have been rationalised and both the smoke tests and the functional tests use the same supplier user now. As a consequence this test is failing because the supplier name is different.


allait:
:+1:

PR #80, opened 2015-09-24
Fix step that was not checking for the correct value

andrewchong:
Previously checked that supplier name text "DM Functional Test Supplier" existed. This was a problem as it was not checking for an exact match. There are two suppliers with name that contained "DM Functional Test Supplier" and the tests were always passing. The change now checks for the exact supplier names:
DM Functional Test Supplier
DM Functional Test Supplier 2


quis:
:+1:

PR #81, opened 2015-09-28
[#102567786] (Potentially) Run functional tests locally.

pcraig3:
This pull request makes it possible to run the functional tests both locally as well as in preview, staging, or production.  Might make more sense to look through it commit by commit.

At the moment, this is only possible by:
- Running all implicated apps locally
- Having [nginx set up locally](https://gist.github.com/netpoetica/5879685)
- Setting a whole bunch of environment variables 

Ways to address these issues: 
- Spinning up apps automatically using something like [Foreman](http://theforeman.org/)
- Running our apps inside containers (with Postgres, Nginx, Python, w/e else) -- we could have nginx installed and configured as we liked it.
- Since the environment variables don't change, there's no reason we can't have them put in a private script somewhere and use flags to decide whether to run tests against our apps locally, in preview/staging/production, or some combination.

It's also necessary, when testing locally, to import the data from preview ([as described in the manual](https://github.gds/pages/gds/digitalmarketplace-manual/developer-setup.html#importing-stuff)).  [I was planning to disregard tests](https://github.com/alphagov/digitalmarketplace-functional-tests/commit/5588dc70709eb05937de0f450bf0eeb951ab0fbb) which assumed the prior existence of data, but @andrewchong's perspective was that not testing for all functionality was the wrong thing to do -- testers should just have preview data imported.

### Fixtures?
Our methods for creating both users and suppliers in the setup steps pretty much override all the defaults set in the fixtures.  With this in mind, [I started by stripping everything out of the fixtures](https://github.com/alphagov/digitalmarketplace-functional-tests/commit/dd8440f241f4b5c235f6220b2bdea62739797605), leaving only the structure of a supplier / user intact, but eventually [I returned most of the contact information to the supplier fixture](https://github.com/alphagov/digitalmarketplace-functional-tests/commit/8de0a0c702a72d19869437d85a938a4d69b4de97) because I figured tests that were looking for phone numbers or email addresses probably had some merit.
As they are now, neither the [supplier fixture](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/pc_if_required_regionalise/fixtures/test-supplier.json) nor the [user fixture](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/pc_if_required_regionalise/fixtures/test-user.json) contains data that could be directly inserted into the API.  Both [suppliers](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/pc_if_required_regionalise/features/step_definitions/setup_steps.rb#L39) and [users](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/pc_if_required_regionalise/features/step_definitions/setup_steps.rb#L107) need passed-in parameters at the point of creation to make them valid.
Are they okay as they are, or do we want less / more in them.

### Known issues

- One local admin test doesn't work because the Mandrill API key locally is different than preview.
 - Specifically, the preview key always says the email was sent successfully whereas the local key always throws a 500-level error.
- The status checks don't work because `_status` endpoints on locally-running apps don't have a value for the `version` key.
- The `ssp_validation.feature` doesn't work because it hasn't ever.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/102567786)


robyoung:
This looks great! Loads of really good tidying up too. I went through by commit so my comments are by commit too.

https://github.com/alphagov/digitalmarketplace-functional-tests/commit/2a87b8ebc828cee08522c392a4228cff8356c35b - This sounds like something that should be more in our control
https://github.com/alphagov/digitalmarketplace-functional-tests/commit/1d1bd0ae03120410d583f164f2507ff6a4d67a5e - This could be solved more robustly by using [`URI::join`](http://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI.html#method-c-join)

pcraig3:
@ https://github.com/alphagov/digitalmarketplace-functional-tests/commit/2a87b8ebc828cee08522c392a4228cff8356c35b - This is set in the [models.py](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L410) file of the API: the default ordering is by lot and then by service name.  We could maybe make this explicit, but I don't know if this is a user need.

robyoung:
I've used this branch in anger and it worked really well. I've made two small additions
- Added a line to the nginx conf to change the tmp directory for uploaded files. The brew default isn't writable by the nobody user.
- Added a bootstrap script to do the steps required to set up nginx on mac.

If you're happy with them then :+1:

PR #84, opened 2015-09-29
set click_link_or_button to be page.click_link_or_button

andrewchong:
This is in attempt to fix intermittent issue encountered whereby the click action does not seem to have registered and tests fail as a result


quis:
:+1:

PR #86, opened 2015-10-02
Add priceMax to service fixtures

robyoung:
Because the JSON schema has recently changed to require numeric strings
rather than numbers and because the services in the database have
numeric values for priceMax and fields only get overwritten when they're
provided; provide the priceMax to override the value in the database.


TheDoubleK:
:+1:

PR #87, opened 2015-10-05
Update for new admin service editing

robyoung:
The admin frontend service editing now uses the same code as the supplier frontend. The changes here fall into a few categories:
- The pricing fields are now all called priceString so the tests must be
  updated to use the more detailed ID.
- The document URLs are no longer displayed in the page so the `href` of their links must be checked.

## Depends on
- [x] https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/173
- [x] https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/132

:sparkles: This has been tested locally (rebased on top of #81) :sparkles:


allait:
:+1:

PR #88, opened 2015-10-05
Fix pagination limit error

robyoung:
If the `DM_PAGINATION_LIMIT` is not provided the function fails because
`nil` cannot be converted to an integer.


allait:
:+1:

PR #89, opened 2015-10-05
Fix smoke test supplier name

robyoung:
This changed in 21a807 but has not been changed on the environments.
Until this is managed by the functional tests I'm reverting this change.


pcraig3:
Could you link to the commit in the message?

Change looks good.

:+1:

PR #92, opened 2015-10-07
Add a script for running the tests

robyoung:
Manages sets of environment variables in a directory. Create a file in `scripts/envs` called `{ environment }.sh` with all the environment variable exports then to use it run `./scripts/run_tests.sh {
environment }`.

### DIFF
--- a/scripts/envs/.gitignore
+++ b/scripts/envs/.gitignore
@@ -1,0 +1,3 @@
+*
+!example.sh
+!.gitignore
--- a/scripts/envs/example.sh
+++ b/scripts/envs/example.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+export DM_API_DOMAIN=${DM_API_DOMAIN:=http://localhost:5000}
+export DM_API_ACCESS_TOKEN=${DM_API_ACCESS_TOKEN:=myToken}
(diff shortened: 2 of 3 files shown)


pcraig3:
Do we care which family of environment variables we're using?
For example, `DM_ADMINISTRATORNAME` (also) became `DM_ADMIN_EMAIL` in [an earlier pull request](https://github.com/alphagov/digitalmarketplace-functional-tests/commit/c61c155f97e1e951b360b2102206ee1f350fc01f), and the [README](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/README.md) was updated to refer only to the new ones. 

/

Seems like two variables are missing from `example.sh`:
 - `DM_FRONTEND_DOMAIN`
 - `DM_SUPPLIER2_USER_EMAIL`

/ 

Finally, do you think it's worth adding a variable to specify only running one `.feature` file?  If I only wanted to run the admin journey, for example, it might come in useful.

robyoung:
Good spot, I've changed them, thanks.

robyoung:
With regard to allowing specific tags / features to be targeted; yes, maybe, I'm not sure of the most convenient way of doing it yet though. I was going to get this in and leave that for a later update.

pcraig3:
:+1:

PR #97, opened 2015-10-21
Fix service order test with new lot_id based sorting

allait:
Related to and should be merged after https://github.com/alphagov/digitalmarketplace-api/pull/273

#### Fix service order test with new lot_id based sorting
Services returned by the API are sorted based on internal lot order.
Full lot name is returned in service.lotName and is now displayed
in service list tables.

#### Update service fixtures with frameworkSlug and lowercase lot
frameworkSlug is required for service creation, while frameworkName
field is always ignored.


andrewchong:
Looks good.
:+1:

PR #98, opened 2015-10-26
Make sure smoke tests for with both old and new API

allait:
While the API lot update is not released to production we need to
make sure smoke tests support both old and new API responses.


TheDoubleK:
:+1:

PR #99, opened 2015-10-26
Get email addresses for supplier users from environment vars

tombye:
You can set the email addresses for the test users through the environment variables in `scripts/envs` so the tests should use those, not be hard-coded.


andrewchong:
You could perhaps incorporate the actual checks on the page presented into your new implementation rather than have it call the existing step and deleted the old implementation all together?

tombye:
@andrewchong I agree and added a fix for that in f7ca82f

andrewchong:
:+1:

PR #104, opened 2015-11-06
Update selector for count of live services

quis:
Fixes failures caused by https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/324


robyoung:
:+1:

PR #105, opened 2015-11-10
Check for Q suppliers for no-pagination checks

TheDoubleK:
When G-Cloud 7 goes live the number of 'D' suppliers may go over 100 and make the tests fail. There are much fewer 'Q' suppliers so the tests are less likely to break using Q.

Looking at this chore: https://www.pivotaltracker.com/story/show/107475586
I think this looks like the only change that (might) be needed for tests to work after G7 is live and G5 expired (I can't see anything specifically to do with G5 services in the tests).


andrewchong:
You are correct. There aren't any tests that are G5 specific.
:+1:

PR #106, opened 2015-11-12
Add feature for admin framework agreements view

robyoung:
## Add feature for admin framework agreements view
This tests the view in the admin app for listing signed framework agreements.

I have tried to keep the step definitions small and put most of the logic into helper functions so that they can be easily reused.

## Match URLs ignoring parameter order
For our applications the order of URL parameters does not matter. Also, because the URLs are generated from dictionaries the order of URL parameters is non-deterministic. This change introduces a couple of new RSpec matchers that compare two URLs based on their component parts, ignoring the order of query or fragment parameters. I have tried to keep the step definitions small and put most of the logic into helper functions so that they can be easily reused.

## Refactor API steps to use call_api
Refacotr existing uses of RestClient to use the recently introduced wrapper call_api. As part of this change call_api was changed so that it can accept a custom domain and parameters.

### DIFF
--- a/features/admin/download_agreements.feature
+++ b/features/admin/download_agreements.feature
@@ -1,0 +1,29 @@
+@not-production @functional-test
+Feature: Admin users can download signed framework agreements
+
+Background:
+  Given I have test suppliers
+  And I have logged in to Digital Marketplace as a 'Administrator' user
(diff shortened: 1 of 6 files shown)


andrewchong:
Looks good.
:+1:

PR #108, opened 2015-11-16
Changed test step to fix failing functional tests

andrewchong:
Supplier A-Z list for suppliers starting with C prior to G7 live had <200 records
G7 live changes this and so the test has been changed to select a supplier(suppliers starting with "T") with <200 records
to fix the failing test


robyoung:
Ran locally against preview, looks good. :+1:

PR #109, opened 2015-11-16
Removed unecessary check for smoke test step

andrewchong:
Test was failing because a check was made for the existence of text on the supplier dashboard page
on the supplier being logged in, which is not necessary and made the test fragile to changes when 
all it needs to do is ensure that the supplier log in was succesful.


robyoung:
:+1:

PR #110, opened 2015-11-23
DOS-Tests for digital outcomes, user research studios and participants service submission

andrewchong:
First attempt at tests for submitting services. Further scenarios to come.
Refactored and made changes to current G-Cloud SSP tests to account for flow changes and DOS service submission, allowing reuse of code between the two frameworks.


features/DOS/ssp_digital_outcomes_service.feature
```diff
@@ -0,0 +1,78 @@
+@not-production @functional-test @ssp
```
TheDoubleK: Should this perhaps be `@ssp-dos`, seeing as we also have `@ssp-g-cloud` now?

features/step_definitions/ssp_steps.rb
```diff
@@ -22,15 +22,23 @@
 end
 
 Given /^I am at the g7 services page$/ do
-  visit "#{dm_frontend_domain}/suppliers/frameworks/g-cloud-7/services"
+  #visit "#{dm_frontend_domain}/suppliers/frameworks/g-cloud-7/services"
```
TheDoubleK: These commented-out lines can probably be deleted if the generalised version is working OK?

features/step_definitions/ssp_steps.rb
```diff
@@ -75,11 +106,19 @@
 end
 
 Then /^I should be on the g7 supplier dashboard page$/ do
```
TheDoubleK: Should this be changed to something like `I should be on the framework dashboard page`, as it is no longer G7-specific.

features/step_definitions/ssp_steps.rb
```diff
@@ -75,11 +106,19 @@
...
-  URI.parse(current_url).path.should == "/suppliers/frameworks/g-cloud-7"
+  #URI.parse(current_url).path.should == "/suppliers/frameworks/g-cloud-7"
+  URI.parse(current_url).path.should == "/suppliers/frameworks/#{store.framework_name}"
 end
 
 Then /^I should be on the g7 services page$/ do
```
TheDoubleK: Also here - something like `I should be on the framework services page`

features/step_definitions/ssp_steps.rb
```diff
@@ -22,15 +22,23 @@
 end
 
 Given /^I am at the g7 services page$/ do
```
TheDoubleK: This is another name that could change if the method isn't only for G7 now.

TheDoubleK:
Few minor comments but otherwise looks good.

andrewchong:
@TheDoubleK: I have made changes as per your comments. Can you please have another look when you are able to.

TheDoubleK:
Looks good.  The only other thing I notice is that `g-cloud_supplier_creation.feature` isn't really G-Cloud specific - it's the same account for whatever framework, so could probably be re-named and moved out of the "G-Cloud" folder.

But it probably doesn't matter much either way, and if you think that that journey belongs with the rest of the G-Cloud tests then that's fine.

:+1: either before or after the move of this feature file, just something to think about.

PR #111, opened 2015-11-30
Fix update_framework_status method to reflect the API change

allait:
Update framework status endpoint now requires the clarification
questions open status to be provided with the framework status.

Adding the argument to the `update_framework_status` with quesiton
status based on framework status (questions are open for 'open'
frameworks and closed for all other framework statuses).


andrewchong:
:+1:

PR #115, opened 2015-12-03
fix for messaging change to draft services

andrewchong:


TheDoubleK:
:+1:  - nice to not need that `servcie_type` switch any more.

PR #116, opened 2015-12-04
Supplier websites are now linked

TheDoubleK:
Update to the supplier app adds hyperlinks to supplier's website fields, so the test need to look inside the anchor tag to find the required text.


robyoung:
:+1:

PR #118, opened 2015-12-04
Updates in tests to match recent content updates

TheDoubleK:
These still won't all pass green until the Preview API has new JSON schemas, but it's a start to getting the tests green.


allait:
:+1:

PR #124, opened 2015-12-10
Log out if already logged in for "Log in" step

TheDoubleK:
Now the login page will redirect to the dashboard if a user is already logged in, so the "Log in" step must log out first in order to complete.

Tested locally and all tests now pass:
screen shot 2015-12-10 at 11 39 49


andrewchong:
Was just looking into this but you have the solution.

andrewchong:
:+1:

TheDoubleK:
Took me longer than it should have to track it down though!

PR #127, opened 2015-12-22
Fix tests for updated admin app

robyoung:
The frameworks repo and frontend toolkit have been upgraded. This has causes the pricing fields and content section URLs to change.


pcraig3:
Seems a shame that three of the inputs say `price` and one of them says `pricing`.

:+1:

PR #130, opened 2016-01-07
Fix for failing tests where link name changed

andrewchong:
Includes check for new link and changed checking for content
to checking for link name 'Log out'


TheDoubleK:
Changes look good  :+1:

PR #131, opened 2016-01-08
Ac new scenarios admin journey 080116

andrewchong:


features/step_definitions/journey_test_steps.rb
```diff
@@ -1471,15 +1472,27 @@ def filter_to_check(filter_name,filter_value,filter_exist)
   page.should have_selector(:xpath, ".//*[@id='global-breadcrumb']/nav/*[@role='breadcrumbs']/li[1]//*[contains(text(), 'Admin home')]")
 end
 
-Then /I am presented with the G-Cloud 7 Statistics page$/ do
-  page.find(
-    :xpath,
-    "//p[contains(text(), 'G-Cloud 7')]/../h1[contains(text(), 'Statistics')]"
-    )
-    current_url.should end_with("#{dm_frontend_domain}/admin/statistics/g-cloud-7")
+Then /I am presented with the '(.*)' page$/ do |statistics_page|
```
TheDoubleK: As this step definition is only for statistics pages it might be better to move the 'statistics' part out of the variable and just have the framework name captured inside the quotes - e.g. 
`/I am presented with the '(.*)' statistics page$/ do |framework_name|`

Because as it is it looks like this step could be used for any type of page in the scenarios, but e.g. it looks like you could do something like `Then I am presented with the 'G-Cloud 7 dashboard' page` or whatever.

features/step_definitions/journey_test_steps.rb
```diff
@@ -1471,15 +1472,27 @@ def filter_to_check(filter_name,filter_value,filter_exist)
...
+  case statistics_page
+  when 'G-Cloud 7 statistics'
+    page.find(
```
TheDoubleK: Could this line come before the `case`, to avoid repeating the xpath logic?
e.g. (if `framework_name` is the parameter passed in)
```
page.find(
       :xpath,
       "//p[contains(text(), framework_name)]/../h1[contains(text(), 'Statistics')]"
       )
```

features/step_definitions/journey_test_steps.rb
```diff
@@ -1488,22 +1501,73 @@ def filter_to_check(filter_name,filter_value,filter_exist)
...
 end
 
+Then /I am presented with the Service updates page$/ do
+  todays_date = Time.new.strftime("%A %d %B %Y")
+  page.find(:xpath,"//p[contains(text(), 'Activity for')]/../h1[contains(text(), '#{todays_date}')]")
+  page.should have_selector(:xpath, "//*/div/label[@for='audit_date'][contains(text(), 'Audit Date')]")
+  page.should have_selector(:xpath, "//*[contains(@id, 'audit_date') and contains(@placeholder, 'eg, 2015-07-23')]")
+  page.should have_selector(:xpath, "//*/div[@class='option-select-label'][contains(text(), 'Show')]")
+  page.find(:xpath, "//*/div[@class='options-container']//label[@for='acknowledged-1']/input[@type='radio']/..").text().should match('All')
+  page.find(:xpath, "//*/div[@class='options-container']//label[@for='acknowledged-2']/input[@type='radio']/..").text().should match('Acknowledge')
```
TheDoubleK: I think these need a `d` on the end:
screen shot 2016-01-08 at 18 04 32

TheDoubleK:
:+1:  Looks good.

pcraig3:
Tested on preview and it works.

:+1:

PR #135, opened 2016-01-22
Small changes to repair existing functional tests

pcraig3:
Made a few small changes to the functional tests (making larger ones as we speak).

##### Capitalise 's' to find the element we're looking for

Notification messages for completed digital services now refer to
thems 'Digital Services', not 'Digital services.'

##### Remove check for next link in service A-Z listing

Checking for a 'next' link on the supplier A-Z page doesn't really tell
us anything about the A-Z page itself.  If we don't have a 'Next, we might still
on an A-Z page.
Removed it because it seemed besides the point.

##### Refactor notification-selector to actually find an element
Also remove a redundant check.

##### 	Stop testing the G7 SSP stuff
G-Cloud-7 applications are long finished, so running the suite of
unit tests (which break by the way) over the G7 SSP seems a bit pointless.


allait:
:+1:

PR #136, opened 2016-01-26
fix failing tests from content change

andrewchong:
To fix issues caused by this change: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/207.


tombye:
:+1:

PR #137, opened 2016-01-27
Setup steps now update the service statuses

pcraig3:
The `create_service` function used for the setup steps will either
create a new service or update it if that service already exists.

Unfortunately, updating a service wouldn't reset the service's status.
Adding a second API call to update services to 'published' so that
tests run as expected.


robyoung:
Create seems like a funny name if it sometimes updates but I guess it doesn't really matter. Looks good. :+1:

PR #138, opened 2016-01-27
Improve test running

robyoung:
## Add script to test services are up
It would be useful to short circuit the functional tests and tell you if some dependent services are not available rather than just showing you screens of fail.

## Make Nginx bootstrapping more robust
Only do actions if required to avoid sudo when not needed. These changes mean that it's less disruptive to put this on the normal test execution path.

- Check if nginx is not installed before trying to install.
- Check if installed nginx config is different before installing.
- Only reload the nginx config if it has changed.

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -10,7 +10,6 @@ bundle install
 if [ "$DM_ENVIRONMENT" = "local" ]; then
   echo -e "\033[0;34mBootstrapping local environment\033[0m"
   ./scripts/create_users.sh "$DM_ENVIRONMENT"
-  exit
 fi
 
 mkdir -p reports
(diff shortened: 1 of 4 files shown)


scripts/run_tests.sh
```diff
@@ -5,6 +5,14 @@ DM_ENVIRONMENT=${1:-local}
 bundle install
 . "./scripts/envs/${DM_ENVIRONMENT}.sh"
 
+./scripts/test_services.sh "$DM_ENVIRONMENT" || exit 1
+
+if [ "$DM_ENVIRONMENT" = "local" ]; then
+  echo -e "\033[0;34mBootstrapping local environment\033[0m"
+  ./scripts/create_users.sh "$DM_ENVIRONMENT"
+  exit
```
pcraig3: I think this line is stopping `./scripts/run_tests.sh` if run locally.

robyoung: Good spot, thanks. I've removed it.

pcraig3:
:+1:

PR #139, opened 2016-01-28
Fix functional tests now that suppliers can remove services

pcraig3:
Suppliers removing their services has changed pretty substantially, both visually and functionally.  As a result, the functional tests need some updates to the content they're looking for and a few new tests.

- Removed the tests about changing a service from 'Public' to 'Private'
- Added a test for a supplier changing a service to 'Removed'
- Relevant tests look for the removed service notification banners where they exist
- Added a new tag to stop the supplier service edit tests from being run (not sure if Jenkins needs updating)
 - In our preview environment, suppliers can remove their services but not edit them

My supplier and admin journeys are working whether on my local environment or on preview.


allait:
:+1:

PR #140, opened 2016-02-03
Change buyer frontend framework slug

robyoung:
It has changed from gcloud to g-cloud to better match how it is used
elsewhere.


andrewchong:
:+1:

PR #141, opened 2016-02-03
Make setup steps user creation more robust

robyoung:
Find users by email address as that is what is unique.
Refactor action functions to reuse found elements.


andrewchong:
Looks good. :+1:

PR #142, opened 2016-02-04
Accept cucumber arguments on run_tests.sh

robyoung:
This allows specific tests or test tags to be run from the `run_tests.sh` and adds some examples.


andrewchong:
Looks good to me.
:+1:

PR #143, opened 2016-02-04
Add pre-flight checks for framework status

robyoung:
In order for the tests to run the frameworks must have particular
statuses. This change adds pre-flight checks to verify them and tell the
user which ones need to be changed.


tombye:
Works like a charm :+1:

PR #146, opened 2016-02-05
AC buyer login

andrewchong:


features/apps_are_up.feature
```diff
@@ -25,7 +25,7 @@ Feature: Apps are up
   @supplier-frontend
   Scenario: Check the supplier frontend is up
     Given I have a URL for "dm_frontend"
-    When I send a GET request to "/suppliers/login"
+    When I send a GET request to "/suppliers"
```
TheDoubleK: Hmm... this is a bit odd, as `/suppliers` will be a redirect (302) to the buyer app login page.  So I'm not sure why this test is passing :confused: - it might be better to point this to a page that doesn't require login though, e.g. `/suppliers/create`

features/supplier/supplier_journey.feature
```diff
@@ -104,3 +104,12 @@ Scenario: Supplier user has 5 failed login attempts and is locked. Login is not
   Given The supplier user 'DM Functional Test Supplier User 3' has 5 failed login attempts
   Then The supplier user 'DM Functional Test Supplier User 3' is 'locked' on the admin Users page
   And The supplier user 'DM Functional Test Supplier User 3' 'can not' login to Digital Marketplace
+
+Scenario: Supplier has forgotten password and requests for a password reset
```
TheDoubleK: This applies to both buyers and suppliers now, but as this is part of the `supplier_journey` steps it's probably fine to leave it as it is.

TheDoubleK:
Apart from the one minor comment this looks good to go :+1: after URL change I reckon.

features/apps_are_up.feature
andrewchong: Noted: Change made accordingly

PR #148, opened 2016-02-10
Changes to fix supplier_framework left in incorrect state

andrewchong:
This change leaves the agreement_returned_at value set at the end of the test so that it correctly reflects a suppliers agrement having been uploaded, rather than resetting it to nothing which it was previously doing


robyoung:
:+1:

PR #16, opened 2015-06-10
Reworked implementation of listing checks on suppliers dashboard and some housekeeping.

andrewchong:
test data clean up will be required on the environments on which these tests have previously already been run or else the tests will fail.
Need to remove the following serviceID:
1122334455667788

PR #149, opened 2016-02-11
Ac ccs admin countersigned agreement

andrewchong:
new ccs/admin scenarios and step implementations-post merge

### DIFF
--- a/features/admin/admin_journey.feature
+++ b/features/admin/admin_journey.feature
@@ -218,3 +218,12 @@ Scenario: As an admin user I want to change the supplier name of a current suppl
   When I change 'input-new_supplier_name' to 'DM Functional Test Supplier 2 name changed'
   And I click 'Save'
   Then I am presented with the 'Suppliers' page with the changed supplier name 'DM Functional Test Supplier 2 name changed' listed on the page
+
+Scenario: As an admin user I want to upload upload Digital Outcomes and Specialists communications
+  Given I have logged in to Digital Marketplace as a 'Administrator' user
+  When I click 'Digital Outcomes and Specialists communications'
+  Then I am presented with the 'Upload Digital Outcomes and Specialists communications' page
+
+  When I choose file 'test.pdf' for 'communication'
+  And I click 'Upload files'
+  Then I am presented with the message 'New communication was uploaded.'
--- a/features/admin/admin_journey.feature
+++ b/features/admin/admin_journey.feature
@@ -194,6 +194,22 @@ Scenario: As an admin user I want to view Service updates
   When I click 'Service updates'
   Then I am presented with the Service updates page
 
+Scenario: As a normal admin user I should not be able to edit a supplier declaration. Link should not exist on the services page.(This is only available to a CCS admin user)
+  Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
+  Then there is no 'G-Cloud 7 declaration' link for any supplier
+  Then there is no 'Digital Outcomes and Specialists declaration' link for any supplier
+
+  When I attempt navigate to the page directly via the URL 'admin/suppliers/11111/edit/declarations/digital-outcomes-and-specialists'
+  Then I am presented with the 'You don’t have permission to perform this action' warning page
+
+Scenario: As a normal admin user I should not be able to upload a contersigned agreement for any supplier. Link should not exist on the services page.(This is only available to a CCS admin user)
+  Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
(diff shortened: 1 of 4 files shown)


features/admin/admin_journey.feature
```diff
@@ -194,6 +194,22 @@ Scenario: As an admin user I want to view Service updates
   When I click 'Service updates'
   Then I am presented with the Service updates page
 
+Scenario: As a normal admin user I should not be able to edit a supplier declaration. Link should not exist on the services page.(This is only available to a CCS admin user)
```
TheDoubleK: `(This is only available to a CCS admin user)` - we have two types of CCS admin user - "category" and "sourcing", but this function is only available to one of them (sourcing I think?) - should we make the exact role explicit in the scenario description?

features/admin/admin_journey.feature
```diff
@@ -194,6 +194,22 @@ Scenario: As an admin user I want to view Service updates
...
+  Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
+  Then there is no 'G-Cloud 7 declaration' link for any supplier
+  Then there is no 'Digital Outcomes and Specialists declaration' link for any supplier
+
+  When I attempt navigate to the page directly via the URL 'admin/suppliers/11111/edit/declarations/digital-outcomes-and-specialists'
+  Then I am presented with the 'You don’t have permission to perform this action' warning page
+
+Scenario: As a normal admin user I should not be able to upload a contersigned agreement for any supplier. Link should not exist on the services page.(This is only available to a CCS admin user)
```
TheDoubleK: Here too - we could be specific about the exact CCS role.

features/admin/ccs_admin_journey.feature
```diff
@@ -0,0 +1,45 @@
...
+  When I enter 'DM Functional Test Supplier' in the 'supplier_name_prefix' field
+  And I click the search button for 'supplier_name_prefix'
+  Then I am presented with the 'Suppliers' page for all suppliers starting with 'DM Functional Test Supplier'
+
+Scenario: As a CCS Sourcing user I should be able to edit a supplier declaration
+    Given I am logged in as a 'CCS Sourcing' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
+    When I click the 'G-Cloud 7 declaration' link for the supplier 'DM Functional Test Supplier'
+    Then I am presented with the admin G-Cloud 7 declaration page
+    When I click the 'Edit' link for 'G-Cloud 7 essentials'
+    And I choose 'No' for 'PR1'
```
TheDoubleK: It might be worth checking the existing value of the fields before making the change - that way the test can be sure that the "Save" button really has saved a new change to the data.  e.g. a new step something like `And 'PR1' has the value 'Yes'` before the `And I choose...` step, and similarly for the `input-SQ3-1k` change.

features/admin/edit_declaration.feature
```diff
@@ -20,6 +20,6 @@ Scenario: As a CCS Sourcing user I should be able to edit a supplier declaration
     And I click 'Save and return to summary'
     Then I am presented with the updated admin G-Cloud 7 declaration page
 
-Scenario: As a normal admin user I should not be able to edit a supplier declaration
-    Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
-    Then there is no 'G-Cloud 7 declaration' link for any supplier
+#Scenario: As a normal admin user I should not be able to edit a supplier declaration
```
TheDoubleK: These comments can be removed now the scenario has moved to `admin_journey`.

TheDoubleK:
Couple of very minor comments, but once they are fixed I think this is good to go :+1:

features/admin/edit_declaration.feature
andrewchong: Removed accordingly

features/admin/admin_journey.feature
andrewchong: Changed to "CCS sourcing admin"

features/admin/admin_journey.feature
andrewchong: Changed to "CCS sourcing admin"

features/admin/ccs_admin_journey.feature
andrewchong: I don't think there is a need to check before hand as the data is loaded from JSON prior to the test being run and should never change, and also the step on line #16 "I am presented with the admin G-Cloud 7 declaration page" already does a check. If that step fails then the scenario fails completely.

andrewchong: If for some reason the setup failed to re-load the data before the scenario is run then the data would be incorrect and would be caught by line #16.

TheDoubleK: OK great - PR is good to merge then.

PR #152, opened 2016-02-18
Ac additional admin scenarios and minor changes

andrewchong:


TheDoubleK:
It would be nice to squash the WIP commits into a single commit for these changes. (`git rebase -i` and then squash/fixup - I can help with this if you're not sure about it at all.)

features/admin/admin_journey.feature
```diff
@@ -47,12 +49,27 @@ Scenario: As an admin user who has logged in to Digital Marketplace, I wish to s
...
+Scenario: As an admin user who has logged in to Digital Marketplace, I wish to navigate to the Supplier Users page via the supplier(s) by supplier name prefix page
+  Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
+  When I click the 'Users' link for the supplier 'DM Functional Test Supplier'
+  Then I am presented with the 'Users' page for the supplier 'DM Functional Test Supplier'
+
+Scenario: As an admin user who has logged in to Digital Marketplace, I wish to navigate to the Supplier Services page via the supplier(s) by supplier name prefix page
+  Given I am logged in as a 'Administrator' and navigated to the 'Suppliers' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
+  When I click the 'Services' link for the supplier 'DM Functional Test Supplier'
+  Then I am presented with the 'Services' page for the supplier 'DM Functional Test Supplier'
+  And I can see all listings ordered by lot name followed by listing name
```
TheDoubleK: Should this be `I can see all services belonging to that supplier...` - just to be clear it is not *all * listings. (I think `services` is a better name than `listings` here, although stick with `listings` if it's used everywhere else.

andrewchong:
I'll grab you tomorrow when I am back in the office to get help with squashing the WIP commits. My day off today and I could not stay away from work. :)

features/admin/ccs_category_admin_journey.feature
```diff
@@ -0,0 +1,80 @@
...
+Scenario: As a CCS Category user who has logged in to Digital Marketplace, I wish to search for services by supplier ID
+  Given I have logged in to Digital Marketplace as a 'CCS Category' user
+  When I enter '11111' in the 'supplier_id_for_services' field
+  And I click the search button for 'supplier_id_for_services'
+  Then I am presented with the 'Services' page for the supplier 'DM Functional Test Supplier'
+  And I can see all listings ordered by lot name followed by listing name
+
+Scenario: As a CCS Category user who has logged in to Digital Marketplace, I wish to search for services by supplier ID and view a specific service
+  Given I am logged in as a 'CCS Category' and navigated to the 'Services' page by searching on supplier ID '11111'
+  When I select 'the' second listing on the page
```
TheDoubleK: Having `the` as the value passed in to the function seems odd to me. Maybe the step definition could change to take something like `when I click the...` and have 'view service link' or 'service name' as possible values?  That would be more descriptive of what it's doing than just `the`.

features/step_definitions/journey_test_steps.rb
```diff
@@ -34,8 +34,12 @@
   when "Supplier"
     page.fill_in('email_address', :with => dm_supplier_user_email())
     page.fill_in('password', :with => dm_supplier_password())
-  when "CCS Sourcing"
-    page.fill_in('email_address', :with => dm_admin_ccs_sourcing_email())
+  when "CCS Sourcing", "CCS Category"
+    if user_type == "CCS Sourcing"
```
TheDoubleK: I don't see the need for the nested "if" here?  It would be clearer just to have separate `case` switches for each user type I think.  Just because the three admin accounts all use the same password at the moment doesn't mean they always will, so it's probably better to set password separately for each account anyway..

features/admin/admin_journey.feature
andrewchong: "Listings" is used everywhere else. I am leaving this as is

features/admin/ccs_category_admin_journey.feature
andrewchong: You are correct. I have made the change accordingly and it now reads as:
When I click the service name link for the second listing on the page

I removed the need to pass a value in as it's no longer required.

features/step_definitions/journey_test_steps.rb
andrewchong: What was I doing here? It was unnecessary for the nested "if". Removed.

TheDoubleK:
Good stuff :+1:

PR #155, opened 2016-03-01
Ac buyer create brief

andrewchong:


robyoung:
I haven't tested this yet but either way it would be nice to have those wip commits squashed.

robyoung:
Also, a bit of a description would be helpful.

PR #156, opened 2016-03-05
Ac admin scenarios

andrewchong:
- Add new admin scenarios for CCS Category and CCS Sourcing.
- Refactored some steps and did some tidying up


robyoung:
I found it really hard to follow what each commit was addressing. They each seemed to include a few different things. I think in future it would be helpful to split commits so they each address one concept and provide more detailed comments.

Commits from the [data api](https://github.com/alphagov/digitalmarketplace-api/commits/master) may be good examples.
Also the [GDS git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md) provides some good guidelines.

That said, the changes seem good. :+1:

PR #158, opened 2016-03-09
Fix a failure introduced by #156

robyoung:
On the other environments the supplier name is 'Digital Marketplace Team'.

This may not be the correct fix long term but at the moment our production smoke tests are broken.


sarahonetui:
reviewed with @robyoung and tested passing green. :+1:

PR #160, opened 2016-03-09
fix for failing smoke tests

andrewchong:
commented block of link checks that will only be valid when DOS is live


robyoung:
:+1:

PR #163, opened 2016-03-14
Supplier responds to a brief

pcraig3:
**Note:**
This set of functional tests creates 3 new published briefs each time it runs.  Not sure if this is a problem or if we're happy to ignore it.
***

### Create briefs; return users
Updated the `create_and_return_buyer_brief()` function so that it can create a publishable brief and (optionally) publish it.
Returning a user from user creation because we need to be able to associate a published brief with a buyer user.

### Delete non-live briefs
Since the functional tests are constantly creating new briefs, this step deletes those that it's able to.
Live briefs can't be deleted, nor will expired briefs be able to.

### Supplier responds to a brief
- set up supplier users
- set up a buyer user
- publish a brief
- sign in as a supplier
- visit the brief response page (directly)
- check for the correct title and requirements
- answer brief response questions
- submit response
- check for success message

### DIFF
--- a/features/buyer/buyer_create_briefs.feature
+++ b/features/buyer/buyer_create_briefs.feature
@@ -20,7 +20,8 @@ Scenario: Start brief creation for an individual specialist
   When I choose 'North East England' for 'location'
   And I click 'Save and continue'
   Then I should be on the "Overview of work" page for the buyer brief 'Find an individual specialist brief'
-  @wip
+
+@wip
 Scenario: Newly created brief should be listed on the buyer's dashboard and the count of unanswered questions is correct
   Given I am logged in as 'DM Functional Test Buyer User 1' 'Buyer' user and am on the dashboard page
   Then The buyer brief 'Find an individual specialist brief' 'is' listed on the buyer's dashboard
--- a/features/step_definitions/setup_steps.rb
+++ b/features/step_definitions/setup_steps.rb
@@ -3,6 +3,9 @@ require "rubygems"
 require "rest_client"
 require "json"
 require "test/unit"
+require "ostruct"
+
+store = OpenStruct.new
 
 def create_supplier (supplier_id, supplier_name, supplier_description, supplier_contactName, supplier_email, supplier_postcode, supplier_dunsnumber)
   file = File.read("./fixtures/test-supplier.json")
@@ -71,7 +74,7 @@ Given /^The test suppliers have live services$/ do
     create_live_service(11112,"1123456789012354","iaas")
 end
 
-def create_user (email,username,password,role,supplierid=nil)
+def create_and_return_user (email,username,password,role,supplierid=nil)
   file = File.read("./fixtures/test-user.json")
   user_data = JSON.parse(file)
   user_data["users"]["emailAddress"] = email
@@ -85,18 +88,20 @@ def create_user (email,username,password,role,supplierid=nil)
     response = call_api(:post, "/users", payload: user_data)
     response.code.should be(201), response.body
   end
+  user = JSON.parse(response.body)["users"][0]
+  return user
 end
 
 def create_supplier_user (supplierid,email,username,password)
-  create_user(email,username,password,"supplier",supplierid)
+  return create_and_return_user(email,username,password,"supplier",supplierid)
 end
 
 def create_buyer_user (email,username,password)
-  create_user(email,username,password,"buyer")
+  return create_and_return_user(email,username,password,"buyer")
 end
 
 def create_admin_user (email,username,password)
-  create_user(email,username,password,"admin")
+  return create_and_return_user(email,username,password,"admin")
 end
 
 And /^The test suppliers have users$/ do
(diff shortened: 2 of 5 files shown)


robyoung:
Should we be able to delete briefs that have ever been published? Is an expired brief one for an expired framework?

features/step_definitions/setup_steps.rb
```diff
@@ -4,6 +4,10 @@
 require "json"
 require "test/unit"
 
+@buyer_id = nil
+@published_brief = nil
```
robyoung: I don't think this is correct. In Ruby `@` means an instance variable. As far as I can tell this means instance variables on the module.

features/buyer/buyer_create_briefs.feature
```diff
@@ -20,7 +20,8 @@ Scenario: Start brief creation for an individual specialist
   When I choose 'North East England' for 'location'
   And I click 'Save and continue'
   Then I should be on the "Overview of work" page for the buyer brief 'Find an individual specialist brief'
-  @wip
+
+@wip
```
TheDoubleK: I don't think this needs to be `@wip` any more - the buyer dashboard does list briefs now, so this test should run OK:
screen shot 2016-03-15 at 10 49 47

TheDoubleK:
All the test published briefs are going to make it a bit of a mess for anyone trying to use preview for stuff (e.g. user lab sessions).  

Maybe it would be OK to use staging for that, but the problem with staging is that it gets wiped every night, and it'd be nice to be able to set up some data somewhere and have it usable without getting drowned out by functional test noise.

We should be able to clean up the published briefs by calling the API directly to delete I think (might have to switch them back to draft first?) - if we can't do it right now then maybe that's something we should enable in the API.

features/step_definitions/setup_steps.rb
```diff
@@ -71,7 +73,48 @@ def create_live_service (supplier_id, service_id, lot)
...
+    :xpath,
+    "//*[contains(text(), 'Service status')]/following-sibling::*[@class='selection-button selection-button-selected'][text()]"
+  ).text()
+end
+
+def service_status_public (service_id)
+  page.visit("#{dm_frontend_domain}/admin/services/#{service_id}")
+  page.should have_content('Service status')
+  current_service_status = ''
+  while current_service_status != 'Public'
```
TheDoubleK: LOL at this loop.

robyoung:
@TheDoubleK at the moment we can't delete live briefs or move them back to draft by design. I'd be reluctant to add that ability just for the sake of the tests. It implies that it's a valid thing to do, which I don't think it is.

I can see a few alternative approaches we might want to consider.
- Create specific routes in the API for the functional tests that are not available in staging or production
- Have the functional tests delete the data directly from the database :exclamation: :exclamation: 
- Use staging for user research and script the creation of fixture data

TheDoubleK:
Looking at this I think we should be able to move briefs back to draft by calling the API:
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/briefs.py#L138

TheDoubleK:
I don't think scripting fixture data for staging is a good solution - Kevin, Jacob, Catherine, Ralph or whoever should be able to set up the data they need for user research on an ad hoc basis without devs needing to get involved.

robyoung:
Returning a brief to draft after it's been made live seems wrong, is that by design or a bug?

TheDoubleK:
I think it's by design and OK - the API should be able to do what it wants with its data, and updating a status to any allowed value is part of that.  It's up to the frontend apps to enforce what API calls actually get made though.  It's the same principle as checking user roles at the frontend apps - it's a level of "business logic" that the API doesn't need to be aware of.

TheDoubleK:
OK sorry, I'm wrong - it can't be done, as designed by @allait:
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L889

TheDoubleK:
I'm not mad on any of the three options for fixing it though...

features/support/api_helpers.rb
```diff
@@ -12,7 +11,13 @@ def call_api(method, path, options={})
   if payload.nil?
     RestClient.send(method, url, options) {|response, request, result| response}
   else
-    RestClient.send(method, url, payload.to_json, options) {|response, request, result| response}
+    # can't send a payload as part of a DELETE request
```
TheDoubleK: Not strictly true - `# can't send a payload as part of a DELETE request using the ruby rest client` would be more accurate.

TheDoubleK:
After discussion elsewhere we will live with the published briefs.

So the only thing left to do is remove that `@wip` and check that the test passes OK I think.

robyoung:
:sparkles: I've run these tests against preview and can confirm they pass :sparkles:

TheDoubleK:
Including the buyer dashboard one marked as @wip but that shouldn't be?

andrewchong:
Those @wip scenarios are not implemented as yet

TheDoubleK:
Ah OK, sorry.

:+1: to this then.

pcraig3:
Hey @andrewchong, I forgot to replace the word `brief` in the Scenarios, but I'll open a new pull request that does it.

PR #164, opened 2016-03-14
Ac buyer brief

andrewchong:
-Changed test brief creation implementation to use API instead of calling several steps to create a test brief
-House keeping; cleaned up steps by removing those that are unused, renamed a few variables
-Renamed, reworked and added methods for brief tests; copied from Paul's "pc_return_requirements_response" branch
-Added step for brief creation, deletion and navigation to "Overview of Work" page for a newly created brief
-Refactored edit and checking of created draft briefs


features/step_definitions/journey_test_steps.rb
```diff
@@ -1854,10 +1855,10 @@ def check_latest_row_correct(service_status)
     page.should have_link('How to talk to suppliers before you start')
     page.should have_link('how to buy')
     page.should have_button('Choose specialist role')
-    store.service_type = "user-research-participants"
-    current_url.should end_with("#{dm_frontend_domain}/buyers/frameworks/#{store.framework_name}/requirements/#{store.service_type}")
+    store.lot = "user-research-participants"
+    current_url.should end_with("#{dm_frontend_domain}/buyers/frameworks/#{store.framework}/requirements/#{store.lot}")
   when "Find a user research lab"
-    puts "page slow has not been defined/developed"
+    puts "page flow has not been defined/developed"
```
pcraig3: Haha, this is good.

features/step_definitions/journey_test_steps.rb
```diff
@@ -1920,30 +1926,62 @@ def check_latest_row_correct(service_status)
...
+  case item_to_change
+  when "Specialist role", "Location"
+    step "I choose '#{new_value}' for '#{field_name}'"
+  else
+    step "I change '#{field_name}' to '#{new_value}'"
+  end
+
+  steps %Q{
+    And I click 'Save and continue'
+    Then I should be on the "Overview of work" page for the buyer brief 'Find an individual specialist brief-edited'
```
pcraig3: Could you just use the already stored brief name?

features/step_definitions/setup_steps.rb
```diff
@@ -230,19 +237,69 @@ def create_buyer_brief (brief_name,framework_slug,lot,user_id)
...
-  # if response.
-  # if response.code == 404
-  #   response = call_api(:post, "/users", payload: user_data)
-  #   response.code.should be(201), response.body
-  # end
-end
-#@wip-create a buyer brief
-Given /^I have a brief$/ do
-  create_buyer_brief("Individual Specialist-Brief deletion test","digital-outcomes-and-specialists","digital-specialists",10348)
+  brief = create_and_return_buyer_brief("Individual Specialist-Brief deletion test", "digital-outcomes-and-specialists", "digital-specialists", store.buyer_id)
```
pcraig3: I was actually wondering if we could change this name to be a bit more general.
Would there be any problem changing "Individual Specialist-Brief deletion test" to something like "Individual Specialist Brief" or something?

features/step_definitions/ssp_steps.rb
```diff
@@ -101,15 +101,14 @@
...
-
-And /^I uncheck '(.*)' for '(.*)'$/ do |label,field_name|
-  within "##{field_name}" do
-    uncheck label
+    case action
+    when "check"
+      check label
+    when "uncheck"
+      uncheck label
+    end
```
pcraig3: I think you should include defaults for your case statements.  If someone was to write `I 'go to' 'Wales' for 'locations'`, this test still runs green without complaining.

features/step_definitions/setup_steps.rb
andrewchong: Yes, it can. I will change it as suggested.

features/step_definitions/ssp_steps.rb
andrewchong: Added defaults to case statements as suggested.

pcraig3:
Just as a final point, some of the `{something}Locations` keys were changed to `locations` in the past little while to make them simpler, so anything referring to the old one will start throwing errors.

This is where it happened --> https://github.com/alphagov/digitalmarketplace-api/pull/356/files

To fix this, just do a Find+Replace for `outcomesLocations` (which becomes `locations`) and `recruitLocations` (which also becomes `locations`)

pcraig3:
:+1:

TheDoubleK:
I don't like to jump in after the thumbs-up, but we are deliberately not using the word "brief" in any of our user-facing stuff.

As the functional tests are user journeys maybe we shouldn't use the word here either? Might be better to call them "requirements" for buyer journeys and "opportunities" for supplier journeys, as they are in the frontend apps.

andrewchong:
Valid point Kevin. We should keep the language consistent. Have discussed with Rob and agree to change the language in the scenarios but leave it as "brief" in the implementation itself as we already use "brief" in the API. I will make the change accordingly.

robyoung:
:sparkles: this has been run against preview :sparkles:

robyoung:
:+1:

andrewchong:
:+1:

PR #166, opened 2016-03-15
Buyer email account must be a public sector email address

TheDoubleK:
The buyer creation test is failing now that email address is checked up-front rather than once the email link is clicked.

This address will pass the validation check OK.


robyoung:
:+1:

PR #169, opened 2016-03-21
Make sure supplier is eligible before responding to a brief

allait:
### Make sure supplier is eligible before responding to a brief
Creates a live digital specialists service during the brief responses setup.

Removes the service listing check for missing 9th service, since the DOS service might exist depending on when the brief response steps were executed.

### Remove update_details and fix status checks in API helpers steps
API now expects `updated_by` top-level key for write requests. Adds 201 as a valid status when registering supplier framework interest.

### Add framework to create_live_service helper
Adds framework argument to `create_live_service` helper method and renames the files with full framework names.

Removes supplier ID from file names, since the service JSON is the same and we can set supplierId in the helper function.


robyoung:
:sparkles: tested against preview :sparkles: :+1:

PR #170, opened 2016-03-22
Remove JSON file smoke test

allait:
The file has been renamed and the test itself doesn't seem to make
much sense.


robyoung:
:+1:

PR #172, opened 2016-03-23
Update for new link text on "Ask a question" page

TheDoubleK:
If a user is already logged in then they now see the link text "Ask a question" rather than "Log in to ask a question", which is only shown when no user is logged in.


allait:
:+1:

PR #173, opened 2016-03-29
add check for 'closed' brief status in delete brief method

andrewchong:
This is to stop attempted deletion of closed briefs


pcraig3:
:+1:

PR #174, opened 2016-03-29
new scenarios for checking count of unanswered questions

andrewchong:
Check count of required and optional questions left to be completed as presented on the buyer dashboard for the newly created brief

### DIFF
--- a/features/step_definitions/journey_test_steps.rb
+++ b/features/step_definitions/journey_test_steps.rb
@@ -1820,7 +1820,7 @@ And /There is no agreement available on the page$/ do
 end
 
 Given /^I navigate directly to the page '(.*)'$/ do |url|
-  page.visit("#{dm_frontend_domain}#{url}")
+  page.visit("#{dm_frontend_domain}/#{url}")
 end
 
 Then /I am on the '(.*)' page$/ do |page_name|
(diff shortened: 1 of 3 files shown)


features/step_definitions/setup_steps.rb
```diff
@@ -280,7 +280,7 @@ def delete_all_draft_briefs (user_id)
     brief_id = brief["id"]
     updated_by = {updated_by: "Functional tests"}
 
-    if brief["status"] != "live"
+    if !["live", "closed"].include? brief["status"]
```
robyoung: I think this would be more idiomatic as `unless ["live", "closed"].include? brief["status"]`

features/step_definitions/journey_test_steps.rb
```diff
@@ -2052,3 +2057,21 @@ def check_latest_row_correct(service_status)
...
+    if store.required_question_count >= 1
+      page.find(:xpath, "//a[contains(@href, '#{store.current_brief_id}')]/../../../td[3]/span[text()]").should have_content("#{store.required_question_count}"' required')
+    end
+  else
+    page.should have_no_selector(:xpath, "//a[contains(@href, '#{store.current_brief_id}')]/../../../td[3]/span[text()]")
+  end
+end
+
+Given /^I navigate directly to the '(.*)' dasboard page$/ do |user_type|
+  step "I navigate directly to the page \'#{user_type}\'"
```
robyoung: Why do we need this as a separate step? Why can't we just use `Given I navigate directly to the page 'buyers'`?

robyoung:
:+1:

PR #178, opened 2016-03-31
Added test scenario for public view of opportunity 

andrewchong:


features/public/public_view_published_requirements.feature
```diff
@@ -0,0 +1,30 @@
...
+
+    And Summary row 'Start date' 'should' contain '31/12/2016'
+    And Summary row 'Contract length' 'should' contain '1 day'
+    #Public view currently displays 'Location' twice. Reinstate after issue resolved.
+    #And Summary row 'Location' 'should' contain 'Scotland'
+
+    And Summary row 'Background information' 'should' contain 'Make a flappy bird clone except where the bird drives very safely'
+    And Summary row 'Important dates' 'should' contain 'Yesterday'
+
+    And Summary row 'Essential requirements' 'should' contain 'Can you do coding?'
```
robyoung: Quoting the verbs in these statements makes them read strangely. Could this be `Add Summary row 'Essential requirements' should contain 'Can you do coding?'`?

This may just be me.

andrewchong: I see what you mean, but they are in quotes as I am using them as a parameter in the implementation.

robyoung: Yea, I get that. Parameters don't need to be quoted and the convention makes some of the statments read strangely when verbs are quoted. Like I say, it may just be me. If the convention helps I don't think it's a big deal.

robyoung:
:+1:

PR #182, opened 2016-04-07
Buyer public q and a scenarios and implementation

andrewchong:
Minor refactoring to accommodate the reuse of implementation steps

### DIFF
--- a/features/step_definitions/journey_test_steps.rb
+++ b/features/step_definitions/journey_test_steps.rb
@@ -2084,10 +2084,10 @@ Then /^I am taken back to the "Overview of work" page with the (.*) question and
   case question_number
   when 'first'
     store.question_count = 1
-   page.should have_selector(:xpath, ".//div/h2[contains(text(), 'Clarification questions')]/..//tbody/tr[#{store.question_count}]/td[1]//span[contains(text(), '#{store.question_count}.')]/../../span[contains(text(),'#{@value_of_interest['question']}')]")
-   page.should have_selector(:xpath, ".//div/h2[contains(text(), 'Clarification questions')]/..//tbody/tr[#{store.question_count}]/td[2]//span[contains(text(),'#{@value_of_interest['answer']}')]")
-   store.question_1 = @value_of_interest['question']
-   store.answer_1 = @value_of_interest['answer']
+    page.should have_selector(:xpath, ".//div/h2[contains(text(), 'Clarification questions')]/..//tbody/tr[#{store.question_count}]/td[1]//span[contains(text(), '#{store.question_count}.')]/../../span[contains(text(),'#{@value_of_interest['question']}')]")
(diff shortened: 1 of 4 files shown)


features/step_definitions/journey_test_steps.rb
```diff
@@ -2074,8 +2084,51 @@ def check_latest_row_correct(service_status)
...
   page.should have_content(name)
 end
+
+Then /^I am taken back to the "Overview of work" page with the (.*) question and answer listed under "Clarification questions"$/ do |question_number|
+  page.should have_selector(:xpath, ".//div[@class='marketplace-paragraph']/h2[contains(text(), 'Overview of work')]")
+  current_url.should end_with("#{dm_frontend_domain}/buyers/frameworks/#{store.framework}/requirements/#{store.lot}/#{store.current_brief_id}#clarification-questions")
+
+  case question_number
+  when 'first'
+    store.question_count = 1
```
TheDoubleK: Extra spacing here at the start of the line, but not really a problem.

features/step_definitions/journey_test_steps.rb
```diff
@@ -2074,8 +2084,51 @@ def check_latest_row_correct(service_status)
...
+  else
+    fail("The value provided is not valid: #{question_number}")
+  end
+
+  page.all(:xpath, "//div/h2[text()][contains(text(), 'Clarification questions')]/..//tbody/tr").length.should == store.question_count
+end
+
+Then /^I am taken to the list of "Supplier opportunities" page$/ do
+  page.should have_selector(:xpath, "//h1[contains(text(), 'Supplier opportunities')]")
+  page.should have_selector(:xpath, ".//*[@id='global-breadcrumb']/nav/*[@role='breadcrumbs']/li[1]//*[contains(text(), 'Digital Marketplace')]")
```
TheDoubleK: I'm not sure that checking this breadcrumb is a useful test. I don't think that if in future we change the breadcrumbs on the page that should be something that breaks the tests - I reckon just testing the heading is right would be enough here.

TheDoubleK: I think this is a really good general purpose test that we've got to the right page, I think this is all that needs to be checked for many pages - maybe re-use this instead of making a specific `I am taken to the list of` test:
https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/step_definitions/ssp_steps.rb#L139

TheDoubleK:
Reviewed - couple of minor comments but OK to merge as-is :+1:

features/step_definitions/journey_test_steps.rb
andrewchong: removed extra spacing.

features/step_definitions/journey_test_steps.rb
andrewchong: Agree that there is no need to check for breadcrumb-removed.
Applied suggested change to re-use existing page check implementation.

PR #185, opened 2016-04-21
Update scenarios for the new brief overview page

allait:
Rewrites brief creation to use the new brief overview page.
There are a lot of changes that went in in the last few days, so
it's likely I've missed something, but the major changes are:

* There's no brief unanswered question count checks since the
  counters aren't displayed on the overview page anymore and
  it's hard to count questions without a single summary view.

* Removed steps that edited all brief fields. Only brief title is
  changed to check that editing works, other steps don't add that
  much compared to filling brief fields for the first time.

* Moved some step definitions to `requirements_steps.rb`. This is
  complicated by the use of a file-wide global `store` in the
  journey steps.

This will need some clean up in the future, since I've probably left
some unused step definitions behind.

### DIFF
--- a/features/buyer/buyer_create_requirements.feature
+++ b/features/buyer/buyer_create_requirements.feature
@@ -1,4 +1,4 @@
-@not-production @functional-test
+@not-production @functional-test @write-brief
 Feature: Buyer create buyer requirements
 
 Background: Login to Digital Marketplace as the newly created buyer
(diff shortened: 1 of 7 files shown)


allait:
Ran against preview, pass after renaming 'start date' field ✨

TheDoubleK:
Looks good as far as I can see, let's try it and see preview go red again ;) 
:+1:

PR #192, opened 2016-05-03
reinstate view specific service scenario

andrewchong:
scenario for admin user to view specific service via supplier name prefix search route
and clicking on a specific service


TheDoubleK:
Looks good :+1:

PR #195, opened 2016-05-11
Fix for no remove from supplier button

TheDoubleK:
This is branched from: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/194 which should get merged first.

This change needs to be merged just before:
 - [ ] https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/195


allait:
👍

PR #200, opened 2016-06-10
Fix mandatory assessment methods field

allait:
Since field has a default value of 'Work history' functional tests have to use the 'Edit' link to add additional assessment methods.


pcraig3:
👍

PR #202, opened 2016-06-13
Add a check for Q&A session being ticked once complete

TheDoubleK:
New logic ticks this if Q&A session details are added, so nice to have a test for it.

Step passes OK against preview:
screen shot 2016-06-13 at 13 13 52


pcraig3:
Works for me 👍

PR #203, opened 2016-06-13
Added requirement to use rbenv to manage ruby versions

idavidmcdonald:
Requirement added as a solution to Nokogiri installation issues that were run into when installing on Yosemite. Rbenv was needed to be installed before bundler was installed otherwise the native version of Ruby would be used for bundler rather than the rbenv version.


pcraig3:
👍

PR #204, opened 2016-06-14
Fixed supplier creation test caused by updating heading links

idavidmcdonald:
Due to the removal of the 'Create supplier account' link in the heading this test was failing. We have fixed it by changing the link used by this test to start the supplier creation journey using the 'Create a supplier account' link.

[Pull request in buyer app that caused this] (https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/342)


pcraig3:
👍

PR #205, opened 2016-06-14
Revise tests for "Update links in header" story

pcraig3:
The functional tests for logged-in buyers were previously navigating directly to (a) the buyer dashboard and (b) the create buyer account page.  At the point these tests were written, there was probably no other way to visit these pages, but now that there are real user flows the tests can now reflect this.

Was able to remove the `I navigate directly to the page '(.*)'` step completely after revising the above-mentioned tests. 😄 😄 😄 

Needs the following PRs to go in before this can be merged:

- [ ]  https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/494
- [x] https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/343


andrewchong:
Looks good to me. Good to merge when the above mentioned PRs are merged. 👍

PR #209, opened 2016-08-01
Test for open briefs, not live

Wynndow:
This PR is to fix the tests that will break when [this](https://github.com/alphagov/digitalmarketplace-frameworks/pull/296) and [this](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/370) get merged.

Briefs have been changed to being open instead of live. This updates
the tests to reflect that.


andrewchong:
👍

PR #211, opened 2016-08-31
Switch over status route with publish route

pcraig3:
We've deprecated the `/status` route in our API for a more flexible `/publish` route, but the functional tests were still using the old one.

This fixes them.


Wynndow:
👍

pcraig3:
Boom!  Don't even need your help, @idavidmcdonald.

PR #215, opened 2016-10-18
Restore app status tests

allait:
### Update gems and .ruby-version
Some of the packages were few major versions behind the upstream, which makes finding correct documentation much harder.

Updating all our dependencies to the latest versions doesn't seem to break anything.

Also removing gem versions from Gemfile. We don't really have a known reason to avoid major versions of any packages and since we commit Gemfile.lock, bundler will use the pinned versions anyway. Removing versions from Gemfile allows us to use `bundle update` to get the latest available version of any package.

New version of rack requires ruby > 2.2.2, so updating .ruby-version to the latest stable ruby.

### Save capybara screenshots to reports/
`rm screenshot*` can misbehave when there are a lot of screenshots in the folder, so moving all files to `reports/` to make the clean up step a bit simpler.

### Silence rspec warning by allowing `.should` syntax
Even though `expect()` seems like a better approach we're already using `should` in quite a few places and there doesn't seem to be a big benefit in updating at the moment.

### Delete apps_are_up.feature
'Apps are up' check is performed when checking the `/_status` for each app, so there's no need for a separate feature.

### Make MAYBE_VAR a constant
Constants have a different scope rules from local variables and raise a warning if they're redefined.

Removes variables used only by the Transform step from global scope.

### Clean up and combine .gitignore files
Removes all python-related items from .gitignore and merges scripts/envs/.gitignore into the main one. Directory-level .gitignore files are rarely used and so can be a bit unexpected.

### Remove insecure SSL phantomjs options
We have correctly signed certificates in all environments, so there's no reason to silence certificate errors.

### Restore app /_status tests to display current app versions
Uses rest-client to send requests to the /_status pages, since capybara tries to wrap the response with HTML, even for `page.source`, which breaks `JSON.parse`.

To avoid a special step definition for status pages adds a "display"
step to print out application version.

### Add cucumber config with a default profile
Allows running cucumber without any options to get the default report output and show application versions in the first scenario.

https://github.com/cucumber/cucumber/wiki/cucumber.yml

### Replace production user hooks with explicit steps
Adds explicit "I have a production <role> user" step definition that creates the user based on environment variables.

Replaces `_USER_NAME` variables with random names. Names are set when the user is created for the first time.

Removes `support/env` wrappers for `dm_production_` environment variables since they're only used by one step and are easier to write using dynamic role strings.

Restores `I have a user` step for creating one-off user accounts. The step should be used by the functional steps and captures a lot of good-practice we agreed on during the functional mob.

### Replace scripts/ with Makefile and config/ environment scripts
Removes create_users and test_dependencies scripts since their steps should be performed by the test scenarios themselves. App status check are always run first and they check that apps are available. Same should be done for the framework status checks for the steps that require specific ones.

Replacing run_tests script with a Makefile since that's what we're using in other apps. Makefile makes it a bit harder to run a custom set of tags, but it makes the common thing easy, which is the main problem we had in the past.

### Update README
Adds instructions for using new make steps. Removes environment variables list since it's out of date and duplicates the config/ file contents.


pcraig3:
Looks sensible + worked on first try 👍

PR #218, opened 2016-10-24
Add an option to quote the value in MAYBE_VAR

allait:
When searching for service names we need to enter the value in quotes
since individual words can match multiple services. Also special characters
like "&" and ":" are treated as separate terms that don't match anything
since the original field is analyzed by Elasticsearch. Quoting special
characters with other parts of the query solves the issue.

Adding an optional "quoted" match to the MAYBE_VAR allows us to
return a quoted value. This also changes MAYBE_VAR regexp to use
named groups since the transform only cares about certain matches.

Another approach we've considered was adding a separate step for
quoting a stored value:

    And I have a random g-cloud service from the API
    And I wrap that service.serviceName with quotes
    When I enter that service.serviceName in the 'q' field

The problem with this approach is that MAYBE_VAR transform doesn't
return the name of the instance variable or the attributed, only
the value, so the step can't replace the attributed value with a
quoted one. We could return a custom object that contains that
information from the transform, but the current approach seems
simpler.

### Remove `sleep 1` after button click
We should wait for pages to load based on content, not time delays, so following the page transition step with a "I am on a page" step shouldn't require any waiting.

### DIFF
--- a/features/buyer/catalogue.feature
+++ b/features/buyer/catalogue.feature
@@ -50,10 +50,10 @@ Scenario: User is able to search by service id and have result returned.
 Scenario: User is able to search by service name and have result returned.
   Given I am on the /g-cloud page
   And I have a random g-cloud service from the API
-  When I enter that service.serviceName in the 'q' field
+  When I enter that quoted service.serviceName in the 'q' field
   And I click 'Show services'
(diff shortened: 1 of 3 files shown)


pcraig3:
"Looked at & I think it deserves a :+1:"

PR #219, opened 2016-10-24
Reverse result count check to get a more useful error message

allait:
Calling `.should` on the expected value returns a confusing error
message:

    expected: >= 120
        got:    119 (RSpec::Expectations::ExpectationNotMetError)


pcraig3:
👍

PR #220, opened 2016-10-28
Fix "supplier name changed" test to match new redirect location

TheDoubleK:
Change reflects recent admin frontend redirect change.

Tested against preview and runs green:
screen shot 2016-10-28 at 12 17 48


allait:
👍

PR #221, opened 2016-10-31
Remove steps that rely on unsetting framework agreements

TheDoubleK:
Now we're using the new FrameworkAgreement routes we have no way to unset a framework agreement as having been returned, and agreements can no longer be set with a single API call.

This pull request: 
 1. Removes the steps that attempt to unset framework agreements and test the "empty list" of agreements as this is no longer possible. 

 2. Updates the setting of framework agreements to use the new API routes, replacing the old `update_framework_agreement_status` API call with `create_and_sign_framework_agreement` that performs the three API calls now needed to return a signed agreement.

 3. Removes steps that check the existence of agreement files in S3 - the tests don't upload any files, so they shouldn't be testing that they can download one.  These tests previously worked because of the standard S3 file path we used to have for agreement files, and the (coincidental) existence of these files in S3 for the preview test supplier accounts.

### DIFF
--- a/features/admin/admin_journey.feature
+++ b/features/admin/admin_journey.feature
@@ -8,8 +8,6 @@ Scenario: Setup for tests
   And Test supplier users are active
   And Test supplier users are not locked
   And The user 'DM Functional Test Supplier User 3' is locked
-  And no 'g-cloud-7' framework agreements exist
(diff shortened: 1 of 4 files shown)


allait:
👍

PR #225, opened 2016-11-07
Admin links update

TheDoubleK:
Fixes for updated admin app links.

Also reinstates checks on agreement ordering that should now pass OK.


andrewchong:
👍 Looks good to me.

PR #227, opened 2016-11-10
Remove paragraph tags from brief response questions

allait:
Since new questions aren't being processed with markdown there's no paragraph tag inside a span.


Wynndow:
👍

PR #228, opened 2016-11-11
Fix brief response requirements check

allait:
Without a paragraph tag requirements selector matches both
individual requirements and "Do you have the essential skills and
experience?" section titles, which breaks the index-based iteration.

Making the selector more specific should fix the issue.


TheDoubleK:
Fingers crossed... :+1:

pcraig3:
*woop woop* 

👍

PR #229, opened 2016-11-22
Rerun and random failures

allait:
### Check API status code before parsing the response
Gives a more useful error message than 'JSON must have at least two octets'.

### Store failed cucumber scenarios and add a make step to rerun them
Adding an additional output to the default cucumber profile allows us to store the failed scenario names and to rerun them using `make rerun` step. Rerun will overwrite the failed scenarios file, so it can be run repeatedly to only execute scenarios that keep failing.

### Replace generic h1 find with have_selector with text filter
`find('h1')` causes a race condition since it can succeed on any page without waiting for the new response to be displayed. Rewriting it with have_selector and `text:` filter should force Capybara to wait until an h1 element with the correct text is present on the page.

### Escape HTML when printing messages with user input
Cucumber HTML formatter doesn't escape HTML tags in printed messages, so if there are any (and there are since we've created quite a few when testing the apps for XSS) in the random brief/service titles they get injected into the report page.

I've made a PR to the cucumber-ruby, but for now I'm adding the manual escaping. This affects the CLI output, since it will go through the same escaping as well.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -29,6 +29,7 @@ Given /^I have a random g-cloud service from the API$/ do
 
   params = {status: "published", framework: live_g_cloud}
   page_one = call_api(:get, "/services", params: params)
+  page_one.code.should == 200
   last_page_url = JSON.parse(page_one.body)['links']['last']
   params[:page] = if last_page_url then
                     1 + rand(CGI.parse(URI.parse(last_page_url).query)['page'][0].to_i)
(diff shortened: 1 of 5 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -143,7 +145,7 @@
 end
 
 Then /^I am on #{MAYBE_VAR} page$/ do |page_name|
-  find('h1').text.should == normalize_whitespace(page_name)
+  page.should have_selector('h1', text: 'huh' + normalize_whitespace(page_name))
```
TheDoubleK: Huh?

allait: Oops 😂 I was testing that it actually checks the text and forgot to change it back.

leohemsted:
👍

PR #230, opened 2016-11-29
Create/publish requirements

danielknell:


idavidmcdonald:
Nice work Dan. Didn't get time to share my feedback from our chat earlier today as had to rush off to  a meeting so here it is now.

The flexibility of filling in forms and questions of any type is really cool! If you haven't already, do discuss with Rob as he might be able to use some of it!

I think in general, the form helper module could do with some improved commenting. It's hard to know what a method is doing exactly. It was fine when you took us through in person and gave the explanation of what they do, but this context is missing from the code without having to go through line by line. Something like doc blocks describing inputs/outputs/effects or a few small comments might be useful. Simiarly, method names could be more descriptive, in particular `check_only`, `random_for`, `input_list` don't feel like they mean much, if anything.

Overall though, super cool. I know I'd never of thought to come up with something so generic and flexible so it's got a really good base for the rest of our functional tests.

features/buyer/requirements.feature
```diff
@@ -0,0 +1,187 @@
...
+   And I don't see the 'Description of work' link
+   And I don't see the 'Shortlist and evaluation process' link
+   And I don't see the 'Set how long your requirements will be open for' link
+   And I don't see the 'Describe question and answer session' link
+   And I don't see the 'Review and publish your requirements' link
+
+
+Scenario: Create team to provide an outcome
+  Given I am logged in as a buyer user
+    And I use the following substitutions:
```
allait: I'm not sure if I'm understanding this bit correctly, but as far as I can tell we need substitutions when the input label and summary table value don't match the input value. I'm not sure that a single scenario-wide dictionary is a best solution for this. I _think_ our field values don't clash anywhere, but I don't think that's guaranteed in any way. I also think we might have quite a few of these in other flows.

Is there a way to capture both value AND label of a field when filling it in and then using the label to compare the summary table display? This should remove the need for manual translation. Or am I misunderstanding the issue?

danielknell: I cant think of a way that does not involve having to specify the  values for the form ahead of time...

features/step_definitions/requirements_steps.rb
```diff
@@ -0,0 +1,106 @@
...
+  @substitutions = {}
+  @fields = {}
+end
+
+Given "I have created $type requirement" do |type|
+  page.visit("#{dm_frontend_domain}")
+
+  click_on "Find #{type}"
+
+  assert_selector('h1', text: "Find #{type}")
```
danielknell: have_selector is pretty much a wrapper for assert_selector, ill update calls to it.

features/buyer/requirements.feature
allait: Since this affects radio buttons and checkboxes, can we parse it out of the input we've selected?

danielknell: ok,  this is now pulled out of the DOM on a question by question basis.

code is a bit shonkey but it works. see FormHelper#find_substitutions.

danielknell:
i think i've addressed all the issues left on this now

allait:
Looks good 👍

PR #231, opened 2016-12-01
Skip suppliers without DUNS number

allait:
Some of the suppliers in the preview, staging and production DBs were imported from the old database and don't have a DUNS number.

When tests select a supplier without a DUNS number and run the admin scenario with DUNS search the scenario fails since search returns >100 results. There's no reason for us to handle this case (since it's essentially searching for an empty string), so it seems that the simplest solution is to remove these suppliers from the random pool.


pcraig3:
👍  Let's do it.

PR #232, opened 2016-12-02
renaming region to location

danielknell:
Fixing tests broken after merging:

https://github.com/alphagov/digitalmarketplace-frameworks/pull/340


idavidmcdonald:
I don't think this is what will fix the broken test. This bit of code appears to be unrelated to the failing test. I have opened a new PR that I believe fixes the issues - https://github.com/alphagov/digitalmarketplace-functional-tests/pull/233

PR #234, opened 2016-12-12
Update text to match new content

tombye:
Fixes for changes in https://github.com/alphagov/digitalmarketplace-frameworks/pull/337.

Note: this keeps the 'Set maximum number of specialists you’ll evaluate' link the same as the
old format because it wasn't updated along with the rest of the content. It should be updated once
this is merged (and this link updated when that happened).


idavidmcdonald:
👍

PR #236, opened 2016-12-14
Fix test that uses the empty_message field

tombye:
This field was changed in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/413 (which brought in the changes in https://github.com/alphagov/digitalmarketplace-frameworks/pull/344) so the test needed updating.


idavidmcdonald:
👍 💯

PR #238, opened 2017-01-12
Edit DOS2 declaration rather than G7

TheDoubleK:
The G-Cloud 7 declaration editing is replaced with DOS2 declaration editing here: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/242

This means the functional tests can no longer do their G7 declaration update so this changes to instead edit a DOS2 declaration.

It also, perhaps controversially, removes the test steps that have an admin user upload and download G-Cloud 7 framework agreements.  The G7 links are gone, but it is still a possible flow for DOS agreements. However, with DOS coming to an end in the next month or so I think that whole flow for framework agreement PDFs is effectively obsolete and not worth functionally testing any more.


pcraig3:
Sure, go for it 👍

PR #239, opened 2017-01-23
Delete supplier applying for a brief functional tests

idavidmcdonald:
**This pull request should not be merged until we deploy the new supplier flow to PREVIEW**

[Pivotal story](https://www.pivotaltracker.com/story/show/137359721)

We are deprecating this flow for applying for a brief. As soon as the flow is deprecated on preview then these tests will fail. Therefore we delete these tests.

Corresponding tests will need to be written for the new flow.


TheDoubleK:
Would be nice to drop the new tests in the same time as the old ones go...

idavidmcdonald:
It would be...

They will go be going into separate branches regardless.

TheDoubleK:
OK fair enough.  As long as you promise to write the new ones merge this when the time comes :+1:

PR #244, opened 2017-02-13
Fix all the data

idavidmcdonald:
Fixes two issues with data setup / poor quality assertions due to data being migrated onto preview and breaking the functional tests.


andrewchong:
Looks good to me.

idavidmcdonald:
I'll assume that is a 👍 Andrew....

andrewchong:
Yes, 👍

PR #245, opened 2017-02-13
Hotfix to remove g6 services just in case g6 is live

idavidmcdonald:
However, g6 services are not included in the search index so tests will fail when g6 is live.


TheDoubleK:
Oh this is bad... I can't quite bring myself to approve it... if it was the "jenkins" branch then maybe, but we can't have things like this in our new master suite, can we?

Is there some way to ping the search API to check which frameworks it has in its index? (Possibly not...)

PR #246, opened 2017-02-14
Remove tests that rely on g6 services

idavidmcdonald:
So we can have g6 expired on preview.

Note, the majority of these tests are covered by the new master branch smoke tests.

These tests identified from the following test report when g6 was expired.
https://ci.beta.digitalmarketplace.service.gov.uk/job/functional-tests-preview-legacy/77/functional_test_report/

### DIFF
--- a/features/admin/admin_journey.feature
+++ b/features/admin/admin_journey.feature
@@ -22,20 +22,6 @@ Scenario: As an admin user who has logged in to Digital Marketplace, I wish to s
   And I click the search button for 'service_id'
   Then I am presented with the summary page for that service
 
-Scenario: As an admin user who has logged in to Digital Marketplace, I wish to search for services by DUNS number
-  Given I have logged in to Digital Marketplace as a 'Administrator' user
-  When I enter '930123456789' in the 'supplier_duns_number' field
-  And I click the search button for 'supplier_duns_number'
-  And I am taken to the 'Suppliers' page
-  And I click the 'Services' link
-  Then  I am presented with the 'Services' page for the supplier 'DM Functional Test Supplier'
-  And I can see all listings ordered by lot name followed by listing name
-
-Scenario: As an admin user who has logged in to Digital Marketplace, I wish to search for services by supplier name prefix and view a specific service
-  Given I am logged in as 'Administrator' and navigated to the 'Services' page by searching on suppliers by name prefix 'DM Functional Test Supplier'
-  When I click the service name link for the second listing on the page
-  Then I am presented with the service page for that specific listing
-
 Scenario: As an admin user who has logged in to Digital Marketplace, I wish to search for supplier(s) by supplier name prefix
   Given I have logged in to Digital Marketplace as a 'Administrator' user
   When I enter 'DM Functional Test Supplier' in the 'supplier_name_prefix' field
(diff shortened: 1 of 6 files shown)


features/admin/admin_journey.feature
```diff
@@ -22,20 +22,6 @@ Scenario: As an admin user who has logged in to Digital Marketplace, I wish to s
   And I click the search button for 'service_id'
   Then I am presented with the summary page for that service
 
-Scenario: As an admin user who has logged in to Digital Marketplace, I wish to search for services by DUNS number
```
TheDoubleK: Feels like this wouldn't be so hard to patch up with a G7/G8 supplier?

features/admin/admin_journey.feature
```diff
@@ -22,20 +22,6 @@ Scenario: As an admin user who has logged in to Digital Marketplace, I wish to s
...
-  Given I have logged in to Digital Marketplace as a 'Administrator' user
-  When I enter '930123456789' in the 'supplier_duns_number' field
-  And I click the search button for 'supplier_duns_number'
-  And I am taken to the 'Suppliers' page
-  And I click the 'Services' link
-  Then  I am presented with the 'Services' page for the supplier 'DM Functional Test Supplier'
-  And I can see all listings ordered by lot name followed by listing name
```
TheDoubleK: If we're removing these steps then we should also delete the unused step definitions. (At least, I assume things like  "I can see all listings ordered by lot name followed by listing name" and "I am presented with the service page for that specific listing" are specific to just these steps.)

scripts/test_dependencies.sh
```diff
@@ -29,7 +29,7 @@ echo -e "\033[0;32mAll services are up\033[0m"
 
 test_framework_status "g-cloud-4" "expired"
 test_framework_status "g-cloud-5" "expired"
-test_framework_status "g-cloud-6" "live"
+test_framework_status "g-cloud-6" "expired"
 test_framework_status "g-cloud-7" "live"
 test_framework_status "g-cloud-8" "live"
 test_framework_status "digital-outcomes-and-specialists" "live"
```
TheDoubleK: Do we care about the state of DOS2?

features/admin/admin_journey.feature
idavidmcdonald: This one should be covered by the master tests - https://github.com/alphagov/digitalmarketplace-functional-tests/blob/f5372b4e008e148319ad4366521077a882a90d2c/features/admin/search.feature#L14

pcraig3:
Also, which tests aren't being replaced?  Is the difference between the ones we still cover and the ones we don't based on user types (we have buyer+supplier+admin users in here) or by particular actions (like removing services)?

scripts/test_dependencies.sh
idavidmcdonald: I don't think we have any dependancies on it at the moment... although they may soon come

idavidmcdonald:
Coverage lost roughly appears to be:
- Admin searching by DUNS number and supplier name prefix (we used to have tests for all admin types (admin, category and sourcing). In our new tests we only check with one of these admin types.

- Admin/users changing services from public/private and removing them.

- User searching for services by service ID, service name, service description, keywords field

- Some scenarios regarding navigating to service pages

- Some scenarios about supplier dashboard and viewing their services

features/admin/admin_journey.feature
idavidmcdonald: Thanks, you are right. New commit should deal with removing all unused steps.

TheDoubleK:
Have these run and passed green locally at all? If so it is traditional to add :sparkles: to your pull request.

idavidmcdonald:
Note, these tests are passing when I run them on preview from my own laptop 
image

TheDoubleK:
OK let's do it
:burn::burn::burn::burn:
:+1:
:burn::burn::burn::burn:

PR #247, opened 2017-02-16
Make brief responses agnostic to framework

idavidmcdonald:
At the moment we hardcode 'digital-outcomes-and-specialists' into our code and fixtures. This causes issues as we now move to DOS2. Instead of changing the hardcodings to be DOS2 (which throws it's own set of issues because functional tests run on both preview and staging environments which may have frameworks in different states), we make our tests take the first DOS framework
which is live. This way we don't need to worry about framework states across our environments.

 ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨ 
These tests pass on my local with both DOS1 live and expired.
These tests pass on preview with DOS1 live. 
✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨  ✨

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -3,6 +3,7 @@ Given 'I have a live digital outcomes and specialists framework' do
   response.code.should be(200), _error(response, "Failed getting frameworks")
   frameworks = JSON.parse(response.body)['frameworks']
   frameworks.delete_if {|framework| framework['framework'] != 'digital-outcomes-and-specialists' || framework['status'] != 'live'}
+  frameworks.empty?.should be(false), _error(response, "No live digital outcomes and specialists frameworks found")
   @framework = frameworks[0]
   puts @framework['slug']
 end
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -1,5 +1,14 @@
+Given 'I have a live digital outcomes and specialists framework' do
+  response = call_api(:get, "/frameworks")
+  response.code.should be(200), _error(response, "Failed getting frameworks")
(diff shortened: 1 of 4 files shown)


features/supplier/brief_response.feature
```diff
@@ -2,9 +2,10 @@
 Feature: Supplier applies for a brief
 
 Background:
+  Given I have a live digital outcomes and specialists framework
   Given I have a buyer
   Given I have a supplier
```
idavidmcdonald: Should the last two of these `Given`'s just be `And`'s?

danielknell: Generally you go with one given/when/then & and for any following line.

features/step_definitions/brief_response_steps.rb
```diff
@@ -1,5 +1,14 @@
+Given 'I have a live digital outcomes and specialists framework' do
+  response = call_api(:get, "/frameworks")
+  response.code.should be(200), _error(response, "Failed getting frameworks")
+  frameworks = JSON.parse(response.body)['frameworks']
+  frameworks.delete_if {|framework| framework['framework'] != 'digital-outcomes-and-specialists' || framework['status'] != 'live'}
+  @framework = frameworks[0]
```
danielknell: If this is an empty array?

features/supplier/brief_response.feature
danielknell: Yes,  given/when/then should appear once and following lines should begin with and

idavidmcdonald:
Believe I've addressed your comments @danielknell, let me know :) Thanks

PR #248, opened 2017-02-20
Fix tests for when DOS1 expires

idavidmcdonald:
- Change hardcodings to DOS2 rather than DOS1
- Add fixture and set up test suppliers eligible to apply for DOS2 opportunities

Tests pass on my local with DOS1 expired ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ ✨ 

To be merged after we do the API call to expire DOS1


TheDoubleK:
Looks good :+1:
One question: do we actually have any tests that will depend on DOS1 being expired? If not then we maybe don't need to specify that in the prerequisite checks.

idavidmcdonald:
Fair point. Ran them without that requirement and all is well so will remove

PR #249, opened 2017-02-20
Remove DOS1 framework agreement admin uploading tests

idavidmcdonald:
This behaviour does not exist for DOS2 due to changes in the framework agreement signing process.

As DOS1 is expiring, we will naturally no longer be using this functionality for DOS1 and therefore we remove these tests.


TheDoubleK:
:+1:

PR #250, opened 2017-02-22
Add random string to supplier name

Wynndow:
For this bug on Pivotal: [https://www.pivotaltracker.com/story/show/140326163](https://www.pivotaltracker.com/story/show/140326163)

All suppliers were being created with the same name, which made the
functional tests a little shakey when searching by supplier name.

Adding a random string to the end of the supplier name will make sure
that they're all special snowflakes.

The random string is the same as already used for the suppliers contact
name and email which should make them easy to identify.


idavidmcdonald:
👍

PR #251, opened 2017-02-22
Create users with standard password

Wynndow:
For this chore on Pivotal: [https://www.pivotaltracker.com/story/show/139941307](https://www.pivotaltracker.com/story/show/139941307)

Users were being created with a random password which is unhelpful. Now
we know what it is.

Have run the tests locally and all is well.


idavidmcdonald:
Nicely done 👍

PR #252, opened 2017-02-23
Swap classnames for selected radios and checkboxes

pcraig3:
As mentioned in the CHANGELOG for the frontend toolkit.

Version [20.0.0](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/CHANGELOG.md#2000)

We've changed a classname in our markup so now the tests need to select other things.


idavidmcdonald:
👍

PR #253, opened 2017-02-23
Use class contains rather than explicit string

idavidmcdonald:
So our radio buttons can have additional classes without breaking our tests (e.g. an updated front end toolkit).


pcraig3:
👍

PR #254, opened 2017-03-03
Add tests for suppliers not eligible to apply for a brief

idavidmcdonald:
For this pivotal story - https://www.pivotaltracker.com/story/show/130019195

As we are moving from integration tests for this behaviour in the supplier front end app (see this PR - https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/635 for how and why we are doing this) we add functional tests to cover this behaviour for the main use case

To add the 3 tests for when suppliers aren't eligible this PR:
- adds ability to define the role for a digital specialists service
- adds ability to check a string of text is found at least once anywhere on the page
- adds ability to check if one element is found with an attribute with a certain value. Example usage being 'data-reason' tags which are used by the frontend to send custom dimensions to google
analytics

Tests will fail until this bug fix PR (https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/655) is merged and deployed to PROD.

They will also not be run on jenkins because we are currently excluding all `@brief-response` tests - https://github.gds/gds/digitalmarketplace-jenkins/blob/master/job_definitions/functional_tests.yml#L104
We could change this jenkins job to only exclude the `@eligible-for-brief-response` tests or we could use a different tagging structure like `@skip`. Thoughts?

We might also want to add a test for a supplier trying to apply to a brief they've already applied for (they should be redirected to their response instead).

#### Question
The tests rely on on a DOS2 hardcoding when looking for things such as 'data-reason' attribute values. Is this OK or should we go to the effort now of making them framework independant and taking their value from the `I have a live digital outcomes and specialists framework` step.

#### Question: 
`Then /^I see a '(.*)' attribute with the value '(.*)'/`
or
`Then /^I see an element with a '(.*)' attribute with the value '(.*)'/ `

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -33,8 +33,8 @@ Given 'that supplier is on that framework' do
   submit_supplier_declaration(@framework['slug'], @supplier["id"], {})
 end
 
-Given /^that supplier has a service on the (.*) lot$/ do |lot_slug|
-  @service = create_live_service(@framework['slug'], lot_slug, @supplier["id"])
+Given /^that supplier has a service on the (.*) lot(?: for the (.*) role)?$/ do |lot_slug, role_type|
+  @service = create_live_service(@framework['slug'], lot_slug, @supplier["id"], role_type)
 end
 
 Given 'that supplier has a user' do
--- a/features/supplier/brief_response.feature
+++ b/features/supplier/brief_response.feature
@@ -3,14 +3,47 @@ Feature: Supplier applies for a brief
 
 Background:
   Given I have a live digital outcomes and specialists framework
-    And I have a buyer
-    And I have a supplier
-    And that supplier is on that framework
-    And that supplier has a user
-    And that supplier user is logged in
+  And I have a buyer
+  And I have a live digital-specialists brief
+  And I have a supplier
+  And that supplier has a user
+  And that supplier user is logged in
 
+@not-eligible-for-brief-response
+Scenario: Supplier is not eligible as they are not on the framework
+  Given I go to that brief page
+  And I go to that brief page
+  And I click 'Apply'
+  Then I am on 'You can’t apply for this opportunity' page
+  And I see 'You can’t apply for this opportunity because you’re not a Digital Outcomes and Specialists 2 supplier.' text on the page
+  And I see a 'data-reason' attribute with the value 'supplier-not-on-digital-outcomes-and-specialists-2'
+
+@not-eligible-for-brief-response
+Scenario: Supplier is not eligible as they are not on the digital-specialists lot
+  Given that supplier is on that framework
+  And that supplier has a service on the digital-outcomes lot
+  And I go to that brief page
+  And I go to that brief page
+  And I click 'Apply'
+  Then I am on 'You can’t apply for this opportunity' page
+  And I see 'You can’t apply for this opportunity because you didn’t say you could provide services in this category when you applied to the Digital Outcomes and Specialists 2 framework.' text on the page
+  And I see a 'data-reason' attribute with the value 'supplier-not-on-lot'
+
+@not-eligible-for-brief-response
+Scenario: Supplier is not eligible as they can not provide the developer role
+  Given that supplier is on that framework
+  And that supplier has a service on the digital-specialists lot for the designer role
+  And I go to that brief page
+  And I go to that brief page
+  And I click 'Apply'
+  Then I am on 'You can’t apply for this opportunity' page
+  And I see 'You can’t apply for this opportunity because you didn’t say you could provide this specialist role when you applied to the Digital Outcomes and Specialists 2 framework.' text on the page
+  And I see a 'data-reason' attribute with the value 'supplier-not-on-role'
+
+@eligible-for-brief-response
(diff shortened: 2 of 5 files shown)


features/step_definitions/brief_response_steps.rb
```diff
@@ -33,8 +33,8 @@
   submit_supplier_declaration(@framework['slug'], @supplier["id"], {})
 end
 
-Given /^that supplier has a service on the (.*) lot$/ do |lot_slug|
```
TheDoubleK: ~~Feels like this version without a role should still be useful to us? e.g. for outcomes, labs, cloud-hosting, whatever?~~

Ah OK, just spotted it will still work without any role specified.

features/supplier/brief_response.feature
```diff
@@ -3,238 +3,279 @@ Feature: Supplier applies for a brief
...
-    And that supplier user is logged in
+  And I have a buyer
+  And I have a live digital-specialists brief
+  And I have a supplier
+  And that supplier has a user
+  And that supplier user is logged in
 
+@not-eligible-for-brief-response
+Scenario: Supplier is not eligible as they are not on the framework
+  Given I go to that brief page
```
TheDoubleK: Don't think we should need the duplicate step here?
```
Given I go to that brief page
  And I go to that brief page
```

features/supplier/brief_response.feature
```diff
@@ -3,238 +3,279 @@ Feature: Supplier applies for a brief
...
-    Then I am on the 'Do you have all the essential skills and experience?' page
+  And I click 'Continue'
+  Then I am on the 'Do you have all the essential skills and experience?' page
   When I choose 'Yes' radio button
-    And I click 'Continue'
-    Then I am on the 'Give evidence of the essential skills and experience' page
+  And I click 'Continue'
+  Then I am on the 'Give evidence of the essential skills and experience' page
   When I enter 'first evidence' in the 'Boil kettle' field
-    And I enter 'second evidence' in the 'Taste tea' field
```
TheDoubleK: I look forward to the next PR where all these whitespace changes are reverted to align the `I`s :smile:

features/support/api_helpers.rb
```diff
@@ -190,7 +190,7 @@ def create_supplier
   JSON.parse(response.body)['suppliers']
 end
 
-def create_live_service(framework_slug, lot_slug, supplier_id)
+def create_live_service(framework_slug, lot_slug, supplier_id, role=nil)
```
TheDoubleK: Feels like a method called `create_live_service` shouldn't be restricted to three DOS lots only.  I guess we flesh it out with G-Cloud and labs as they are necessary though :/

features/support/api_helpers.rb
```diff
@@ -211,6 +211,12 @@ def create_live_service(framework_slug, lot_slug, supplier_id)
       puts 'Lot slug not recoginsed'
   end
 
+  if lot_slug == 'digital-specialists' and role
+    service_data['services']["#{role}Locations".to_sym] = service_data['services'].delete(:developerLocations)
```
TheDoubleK: It's not 100% clear what's going on here?

I'm guessing the `.delete` returns the value that has been deleted as well as removing it from the `service_data`, so the role supplied as an argument gets the values that used to exist for developer?  And developer role gets removed at the same time?

It might be nice to have a comment to that effect (or if I'm wrong then to explain whatever is going on here).  Why do we need to delete the developer role at all?

features/supplier/brief_response.feature
idavidmcdonald: Correct, will remove

features/support/api_helpers.rb
idavidmcdonald: Agreed

features/support/api_helpers.rb
idavidmcdonald: Yes you are entirely correct. Good shout on making this more readable, I will add some comments.

We need to delete the developer role so when we apply for a developer brief we are on the specialists lot but do not have the specialists role (i.e. we just have designer and not developer). This is one of the eligibility tests that I have added.

TheDoubleK:
Nice one, sweet :+1:

Sorry it took so long to get this reviewed :(

idavidmcdonald:
No worries. Thanks Kev :)

TheDoubleK:
Oh, damn, I didn't address your questions.

Q1: In an ideal world I'd say yes let's make them framework independent.  But given that I've approved the PR changes already it's probably a bit late.

Q2: I like the phrasing of the second version more but as the test doesn't actually look for any particular element I think it's fine as it is.

idavidmcdonald:
@pcraig3 I've added some more tests to the brief response feature file, however as we know these are currently turned off. The one's I've added though are to do with ineligibility and so the user will not reach the brief application form and instead will get an error page meaning these tests can be be run and pass. I've therefore split the tests into two tags (@not-eligible-for-brief-response @eligible-for-brief-response) and my plan is to change the jenkins job to exclude the @eligible-for-brief-response tests rather than all @brief-response tests. Sound reasonable? When all the tests are green again we can maybe think about whether we will still need all of the different tags for brief responses.

pcraig3:
Let's `@skip` it, David.

pcraig3:
👍  from me.

PR #255, opened 2017-03-06
DOS2 stats link has been replaced with G-Cloud 9

TheDoubleK:
This should fix failing tests because of admin app link change.


TheDoubleK:
Let's suck it and see.

PR #257, opened 2017-03-17
Replace regularly failing opportunity filtering step

idavidmcdonald:
For this pivotal story - https://www.pivotaltracker.com/story/show/141321125

The `I see that brief in one of the pages that follow from clicking` step has been failing often. This is because it loops through many pages after another until it finds the result and pages may randomly break for example during a deployment. This generally doesn't cause the actual jenkins job to fail because it reruns automatically but does make our tests a bit flakey. Therefore I have removed tests that use this step and added assertions that all the briefs returned on the first page of filtering match the correct status and/or lot.

I decided to check all possible options rather than pick a random status or lot. This makes the tests slightly more rigerous in case the environment only contains briefs of a certain lot or status on the first page (e.g. the first page is entirely open briefs).

### DIFF
--- a/features/supplier/opportunities.feature
+++ b/features/supplier/opportunities.feature
@@ -1,4 +1,4 @@
-@smoke-tests
+@smoke-tests @opportunities
 Feature: Passive opportunity supplier journey
 
 Scenario: User can see the main links on the homepage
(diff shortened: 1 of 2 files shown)


features/supplier/opportunities.feature
```diff
@@ -11,59 +11,36 @@ Scenario: User can click through to opportunities page
...
+  Then I see the number of results has decreased
+  And I see all the opportunities on the page are on the 'Digital outcomes' lot
 
-Scenario: Filtering by status doesn't increase result count
+Scenario: User is able to filter by status
   Given I am on the /digital-outcomes-and-specialists/opportunities page
-  When I note the result_count
-  And I check a random 'status' checkbox
+  And I note the result_count
+  When I check the 'Closed' checkbox
```
TheDoubleK: Will these run after a new brief has been created by tests?  Otherwise there's no guarantee that there will be open briefs.  But generally there will be I guess. (Maybe not on staging?)

TheDoubleK:
These steps look reasonable to me.

My only slight concern is that they are at least partly dependent on previous tests having created open briefs (probably on the specialists lot).  But they should mostly pass in isolation, and are not anywhere near as tightly coupled as our Jenkins branch tests.

idavidmcdonald:
I've updated with a second straw man that does not rely on any existing briefs. I think the new version feels pretty rigorous. Let me know what you think...

idavidmcdonald:
PR now updated with my suggested full solution.

benvand:
:+1:

benvand:
A quick question about the `.rb` stuff...
Is there a difference between 
`x.should be y`
and
`x.should == y`

I'm thinking the first only tests against `Bool` ?

idavidmcdonald:
@benvand The first expression is just one of the build in rspec operators (http://www.relishapp.com/rspec/rspec-expectations/v/3-5/docs/built-in-matchers). It would be equivalent to `x.should == True`.

PR #258, opened 2017-03-29
Get @brief-response test working again

pcraig3:

***

This pull request updates the functional tests so that the `@brief-response` scenario works again.

It does not:
- fix the brief creation test
- provide a general way to fill out fields on a page that has hidden fields
- fix all of the general radio `choose()` or checkbox `check()` functions

I've been working on the last one (there are a few functions in `common_steps.rb` around clicking random things that we're not using, it looks like), but then removed all that code for a follow-up pull request.


idavidmcdonald:
👍 

Nice work Paul. Thanks on the very easy to read commit messages too. It might be good to mention in the PR or one of the commits why we are doing this (because the new frontend toolkit has a bunch of hidden fields).

Is your plan to afterwards fix all of the bullet points you mentioned above or just the last one?

pcraig3:
Plan is to fix them all, @idavidmcdonald.

PR #259, opened 2017-03-30
Refactor selecting radios and checkboxes in common_steps.rb

pcraig3:

***

**Update: pull request #269 takes this code and builds more on top of it.**

What this pull request does:

Whereas my last pull request fixed only the steps needed to get the `@brief-response` tests to work, this one fixes the rest of the steps in this file to do with radios and checkboxes and tries to do it in a way that's reasonably clean.

The generic function for choosing radios and checking checkboxes will accept either a value (as a string) or an input, and then will only select the input if its label is visible onscreen.  Since all of the inputs are hidden, the only way we can make sure that the question isn't totally hidden is by checking its parent label.  Once the label is found, the input will be clicked and its value will be printed as part of the output.

It seems to me this is a reasonable refactor, but if anyone has opinions about where this general code should go, I'm happy to listen.

What this pull request does not do:

- fix the brief creation test
- provide a general way to fill out fields on a page that has hidden fields

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -115,7 +115,9 @@ When /I click #{MAYBE_VAR} ?(button|link)?$/ do |button_link_name, elem_type|
 end
 
 When /I check #{MAYBE_VAR} checkbox$/ do |checkbox_label|
-  page.check(checkbox_label)
+  options = {allow_label_click: true}
+
+  page.check(checkbox_label, options)
 end
 
 When /I choose #{MAYBE_VAR} radio button(?: for the '(.*)' question)?$/ do |checkbox_label, question|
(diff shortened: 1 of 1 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -1,6 +1,50 @@
...
 
+def choose_radio(label_or_radio, find_options={}, choose_options={})
+  if label_or_radio.is_a? Capybara::Node::Element and label_or_radio[:type] == 'radio'
+    radio = label_or_radio
+  else
+    radio = find_field(label_or_radio, find_options.merge({visible: :all}))
+  end
+
+  # If the label for this radio is not visible, it is effectively hidden from the user
+  radio.find_xpath('parent::label')[0].visible?.should be(true), "Expected label for radio button \"#{radio.value}\" to be visible"
```
galund: Do we always lay out our radio buttons inside their labels? I thought I'd also seen we sometimes do the `<label for=X>` thing, in which case this might not find it.

pcraig3: Yeah, they always will be if we're using the macros.  It's possible we're not in some other context, but we should be changing markup like this if we find it.

PR #262, opened 2017-04-04
Check for broken G-Cloud filters.

galund:
relates to https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/468

### DIFF
--- a/features/buyer/catalogue.feature
+++ b/features/buyer/catalogue.feature
@@ -73,3 +73,12 @@ Scenario: User is able to search by keywords field on the search results page to
   Then I see that service.id in the search summary text
   And I see that service.id as the value of the 'q' field
   And I see that service in the search results
+
+Scenario: User is able to click a filter on the search results page to narrow down the results returned
+  Given I am on the /g-cloud page
+  And I have a random g-cloud service from the API
+  When I click that service.lotName
+  And I tick a random filter
+  And I click 'Filter'
+  Then I see that service.lotName in the search summary text
+  And I see a search result
--- a/features/step_definitions/catalogue_steps.rb
+++ b/features/step_definitions/catalogue_steps.rb
@@ -41,3 +41,9 @@ end
 Then(/^I see #{MAYBE_VAR} in the search summary text$/) do |value|
   find(:xpath, "//*[@class='search-summary']/em[1]").text().should == normalize_whitespace(value)
 end
+
+When /^I tick a random filter$/ do
+  checkbox = all(:xpath, "//form[@action='/g-cloud/search']//input[@type='checkbox']").sample
+  page.check(checkbox[:id])
+  puts "Filter ticked: #{checkbox[:name]}"
+end


features/buyer/catalogue.feature
```diff
@@ -73,3 +73,12 @@ Scenario: User is able to search by keywords field on the search results page to
   Then I see that service.id in the search summary text
   And I see that service.id as the value of the 'q' field
   And I see that service in the search results
+
+Scenario: User is able to click a filter on the search results page to narrow down the results returned
+  Given I am on the /g-cloud page
+  And I have a random g-cloud service from the API
```
TheDoubleK: Will this only give us services that are indexed for search?

Because G-Cloud 8 isn't going to be expired, so if it just returns any live g-cloud service then there's a chance that it will be a G8 and the G8 lots will no longer be filterable.

galund: I think it boils down to this:

* we currently have no way to identify a service that should be in the live index, based on any data stored in the API
* which frameworks are currently indexed depends only on the calls made (manually) to the indexing script.

Feels like the only way to fix that would be to introduce a new 'indexed' flag on the frameworks table, and have everything hang off that.

Alternatively we could introduce a new status, i.e. "live-but-not-indexed" (with a pithier name) - but we'd need to review all code currently checking for status "live", so that seems riskier.

features/step_definitions/catalogue_steps.rb
```diff
@@ -41,3 +41,9 @@
 Then(/^I see #{MAYBE_VAR} in the search summary text$/) do |value|
   find(:xpath, "//*[@class='search-summary']/em[1]").text().should == normalize_whitespace(value)
 end
+
+When /^I tick a random filter$/ do
+  checkbox = all(:xpath, "//form[@action='/g-cloud/search']//input[@type='checkbox']").sample
+  page.check(checkbox[:id])
```
pcraig3: I've written some general checkbox checking functions in [my currently open pull request](https://github.com/alphagov/digitalmarketplace-functional-tests/pull/259/files).  Currently:
- it will only exist in the `common_steps.rb` file
- the checkboxes in our option-select boxes haven't been updated to show the new styles (not sure if they will be -- but in its current form, this will break if the markup is updated).

[The general checkbox-checking code](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/68470725021d8d57a8f917fab4b64fb3638de82a/features/step_definitions/common_steps.rb#L18-L30) will work with this though, so if we had a way of sharing it between files, that's probable the safest way to go.

features/buyer/catalogue.feature
galund: Yep, so I'm very happy with the approach in your other PR. I still think we might at some point need to be able to distinguish between 'indexed' frameworks, and 'live but no longer actually in the index' frameworks. But I agree we wouldn't want to introduce something just for the sake of the functional tests.

TheDoubleK:
Is this still pertinent or has it been superseded by more recent changes now?

features/buyer/catalogue.feature
```diff
@@ -73,3 +73,12 @@ Scenario: User is able to search by keywords field on the search results page to
   Then I see that service.id in the search summary text
   And I see that service.id as the value of the 'q' field
   And I see that service in the search results
+
+Scenario: User is able to click a filter on the search results page to narrow down the results returned
+  Given I am on the /g-cloud page
```
pcraig3: I think start directly on the `/g-cloud/search` page rather than starting a page before.

features/buyer/catalogue.feature
```diff
@@ -73,3 +73,12 @@ Scenario: User is able to search by keywords field on the search results page to
...
+  And I have a random g-cloud service from the API
+  When I click that service.lotName
+  And I tick a random filter
```
pcraig3: Small thing, but could we use 'check' instead of 'tick'?  That's how our other tests are worded.

features/step_definitions/catalogue_steps.rb
pcraig3: I think this still stands, but it'll be easier to implement after my new pull request (#269) goes in.

features/step_definitions/catalogue_steps.rb
```diff
@@ -41,3 +41,9 @@
...
+  puts "Filter ticked: #{checkbox[:name]}"
```
pcraig3: "Filter value"?  
[Would make it consistent](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/b2c109e7d4f9a6957f59fc8c0a2c4ae3ee608af5/features/support/helpers.rb#L61) with the other print statements.

features/buyer/catalogue.feature
```diff
@@ -73,3 +73,12 @@ Scenario: User is able to search by keywords field on the search results page to
   Then I see that service.id in the search summary text
   And I see that service.id as the value of the 'q' field
   And I see that service in the search results
+
+Scenario: User is able to click a filter on the search results page to narrow down the results returned
```
pcraig3: This is something we talked about in person, but this scenario name doesn't really describe the thing this test is doing.

features/buyer/catalogue.feature
galund: The issue with 'check' is that in terms of natural language, "I check something" might mean "I go see something is there / working". Maybe only in my head. Happy to change :)

galund:
Will review in https://trello.com/c/5t96gz0Z/400-functional-tests

PR #263, opened 2017-04-10
Fix smoke tests now that G9 is live on preview

pcraig3:
Our buyer smoke tests broke on preview since G9 was set to "live" and the search index was repopulated with entirely G9 services.  We are still keeping G8 "live", however, as ongoing procurements will be referencing some G8 services and we don't want them to say that they're expired.

The things that broke are:
- hardcoded lot names in our buyer smoke tests don't work because the lot names for G9 have changed
- searching for a random g-cloud service from "live" G-Clouds potentially fails if G8 is live but its services aren't indexed
- looking for the search bar on the `/g-cloud` page is broken

After talking with @TheDoubleK, the solutions we've gone with are:
- use a random lot from the latest "live" G-Cloud instead of hardcoded lot names
- select random services from the latest "live" G-Cloud will always find a services that should be in the search index
- look for the search bar on the `/g-cloud/search` page

These changes generalise the functional tests so that they will work across our environments.

### DIFF
--- a/features/buyer/catalogue.feature
+++ b/features/buyer/catalogue.feature
@@ -17,25 +17,12 @@ Scenario: User can click through to g-cloud page
   When I click 'Find cloud technology and support'
   Then I am on the 'Cloud technology and support' page
 
-Scenario: User can get the PaaS search results
+Scenario: User can select a lot from the g-cloud page and see search results.
   Given I am on the /g-cloud page
-  When I click 'Platform as a Service'
+  When I have a random g-cloud lot from the API
+  When I click that lot.name
   Then I am on the 'Search results' page
-  And I see the 'Platform as a Service' breadcrumb
-  And I see a search result
-
-Scenario: User can get the SaaS search results
-  Given I am on the /g-cloud page
-  When I click 'Software as a Service'
-  Then I am on the 'Search results' page
-  And I see the 'Software as a Service' breadcrumb
-  And I see a search result
-
-Scenario: User can get the IaaS search results
-  Given I am on the /g-cloud page
-  When I click 'Infrastructure as a Service'
-  Then I am on the 'Search results' page
-  And I see the 'Infrastructure as a Service' breadcrumb
+  And I see that lot.name breadcrumb
   And I see a search result
 
 Scenario: User is able to search by service id and have result returned.
--- a/features/buyer/catalogue.feature
+++ b/features/buyer/catalogue.feature
@@ -39,33 +39,32 @@ Scenario: User can get the IaaS search results
   And I see a search result
 
 Scenario: User is able to search by service id and have result returned.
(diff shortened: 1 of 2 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -155,7 +167,7 @@
   puts "letter: #{@letter}"
 end
 
-Then /^I see the '(.*)' breadcrumb$/ do |breadcrumb_text|
+Then /^I see #{MAYBE_VAR} breadcrumb$/ do |breadcrumb_text|
```
galund: Must admit I don't follow what the MAYBE_VAR thing is for at all

features/buyer/catalogue.feature
```diff
@@ -19,53 +19,39 @@ Scenario: User can click through to g-cloud page
...
-  And I click 'Show services'
+  And I click 'Filter'
   Then I see that service.id in the search summary text
   And I see that service.id as the value of the 'q' field
   And I see that service in the search results
 
 Scenario: User is able to search by service name and have result returned.
-  Given I am on the /g-cloud page
+  Given I am on the /g-cloud/search page
   And I have a random g-cloud service from the API
```
galund: This is fine, I was wondering whether we could somehow have a test that iterated over at least one service from each of the lots in the latest framework? Trouble with the 'choose at random' approach is, if one of the things breaks (and our lots do all work slightly differently), then we might miss the problem at first.

features/step_definitions/common_steps.rb
pcraig3: `MAYBE_VAR` is some magic thing that @risicle came up with which lets you either use a string value or a variable name in the scenario plain-english-y description.

So before we had a hardcoded breadcrumb value: 
```And I see the 'Platform as a Service' breadcrumb```

But now we have a variable.based one:
```And I see that lot.name breadcrumb```

`MAYBE_VAR` means that both versions will work.  (I think the word 'that' is the switch that decides whether we look for single quotes or do variable introspection.)

features/buyer/catalogue.feature
pcraig3: I dunno. Smoke tests run pretty often and they run on all environments. We only have 3-4 lots, so after they run a few times we're pretty likely to find any issues.

More selfishly, it's not really obvious to me how to make this test do all lots without some really elaborate code, and I personally don't think there's much extra value in checking every lot every time. The tests are pretty basic and don't look for anything besides one search result and one breadcrumb, so they're not really that comprehensive anyway.

TheDoubleK: For the smoketests I think checking one lot per run is fine.  If there are differences between lots then the longer functional test scenarios should cover all these cases.

TheDoubleK:
Looks good to me :+1:

PR #264, opened 2017-04-11
Skip `@skip` tagged tests when running through Makefile

idavidmcdonald:
We have a failing test that we wish to skip running until we have time to fix it. Therefore we tag it with the `@skip` tag and don't run any tests with this tag when running all tests or smoke
tests.

Question.
Previously we use the jenkins jobs to exclude the skip tags (https://github.gds/gds/digitalmarketplace-jenkins/blob/master/job_definitions/functional_tests.yml#L103). Should we continue that pattern and add the skip tag to the smoke tests jenkins job or use this PR and do it on the Makefile?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 DM_ENVIRONMENT ?= local
 
 smoke-tests: setup
-	. config/${DM_ENVIRONMENT}.sh && bundle exec cucumber --tags @smoke-tests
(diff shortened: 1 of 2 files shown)


pcraig3:
I think keeping it in the Jenkins job is better because it means we will always be manually excluding it locally.  What's the argument for doing it in the Makefile?

idavidmcdonald:
Test already fixed by a rogue developer - https://github.com/alphagov/digitalmarketplace-functional-tests/pull/265

PR #265, opened 2017-04-11
Stop checking for lot name for keyword search test

pcraig3:
This test is testing that searching for an id in the search box works, it doesn't actually need to select a lot.
The previous iteration of this test started from the `/g-cloud` page, and clicking the lot would bring you to the `/g-cloud/search` page.  Since the new test is begins on the search results page, we can skip selecting a lot, which was purely navigational.


pcraig3:
Going rogue and pushing this through. 👍 

✨  Tested locally against preview and staging (a couple times each). ✨

PR #269, opened 2017-04-20
Find & select new input fields

pcraig3:

***

What this pull request does:

This pull request fixes the `@requirements` tests and tries to provide some general methods for

- finding all fields on a page
- finding the first field on a page
- choosing a radio
- (un)checking a checkbox

while making sure that the fields are themselves either visible to users or their labels are. 
Once this pull request is merged, we'll have removed all of the remaining `@skip` tags.

This pull request does not: 

- provide any specific help for filling out a page where some inputs are hidden until others are selected (like [this question](https://cloud.githubusercontent.com/assets/2454380/25229940/3ad4cb60-25ca-11e7-9fe7-526b1ab1776b.gif))

If you need that, you can build it yourself -- although it will be easier now that you can find/select inputs without having to check if they're hidden to users or not.

This pull request relies on #259, so either we merge that one first or we merge this one and close that one.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -115,7 +115,9 @@ When /I click #{MAYBE_VAR} ?(button|link)?$/ do |button_link_name, elem_type|
 end
 
 When /I check #{MAYBE_VAR} checkbox$/ do |checkbox_label|
-  page.check(checkbox_label)
+  options = {allow_label_click: true}
+
+  page.check(checkbox_label, options)
 end
 
 When /I choose #{MAYBE_VAR} radio button(?: for the '(.*)' question)?$/ do |checkbox_label, question|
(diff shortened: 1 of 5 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -191,8 +201,8 @@
 end
 
 Then /I see #{MAYBE_VAR} as the value of the '(.*)' field$/ do |value, field|
-  if page.has_field?(field, {type: 'radio'}) or page.has_field?(field, {type: 'checkbox'})
-    page.find_field(field, {checked: true}).value.should == value
+  if page.has_field?(field, {type: 'radio', visible: :all}) or page.has_field?(field, {type: 'checkbox', visible: :all})
+    page.find_field(field, {checked: true, visible: :all}).value.should == value
```
galund: Is it worth having a wrapper around find_field, where `visible: :all` happens automatically? Seems to be a very common pattern anyway.

features/step_definitions/common_steps.rb
```diff
@@ -1,6 +1,18 @@
...
 
+def find_checkboxes_by_name(name, options={})
+  return all_fields(name, options.merge({type: 'checkbox'}))
+end
+
+def find_radios_by_name(name, options={})
+  return all_fields(name, options.merge({type: 'radio'}))
+end
+
+def find_radio_by_name(name, options={})
```
TheDoubleK: Are we likely to ever use this `find_radio_by_name` method at all?  It's not used here, and the approach of finding all radios with a name and checking a random one seems good to me.

TheDoubleK:
Left a question about `find_radio_by_name()` but yep this looks good :+1: with or without it.

pcraig3:
Deleted those three `find_{field}_by_name` methods at the top, @TheDoubleK.

pcraig3:
Merging.

PR #270, opened 2017-04-20
DATEFIELD-fix-for-dates-in-dos2-briefs

benvand:
* Adjust `create_brief` method to return a different breif fixture for dos2 digital outcomes and digital specialists lots.
* Add new fixures to `fixtures.rb` reflecting the changed date format.

http://stackoverflow.com/questions/2134702/ruby-1-9-hash-with-a-dash-in-a-key
https://repl.it/HS9g/0

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -122,15 +122,29 @@ def create_brief(framework_slug, lot_slug, user_id)
   brief_data = {
     updated_by: "functional tests"
   }
-  case lot_slug
-  when 'digital-specialists'
-    brief_data['briefs'] = Fixtures.digital_specialists_brief
-  when 'digital-outcomes'
-    brief_data['briefs'] = Fixtures.digital_outcomes_brief
-  when 'user-research-participants'
-    brief_data['briefs'] = Fixtures.user_research_participants_brief
-  else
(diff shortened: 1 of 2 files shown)


features/support/api_helpers.rb
```diff
@@ -122,15 +122,29 @@ def create_brief(framework_slug, lot_slug, user_id)
...
+  when 'digital-outcomes-and-specialists'
+    case lot_slug
+    when 'digital-specialists'
+      brief_data['briefs'] = Fixtures.digital_specialists_brief
+    when 'digital-outcomes'
+      brief_data['briefs'] = Fixtures.digital_outcomes_brief
+    when 'user-research-participants'
+      brief_data['briefs'] = Fixtures.user_research_participants_brief
+    else
+      puts 'Lot slug not recognised'
```
pcraig3: Could we have these raise exceptions rather than just print something out, or is the syntax more clunky?

benvand: Would rather do that outside the context of this bug fix. It changes the functionality quite dramatically in that it will exit fail instead of passing on an unrecognised lot slug.

pcraig3: Okay, fine with me.

PR #272, opened 2017-04-26
Fix dates

benvand:
* Removed test cases already covered in:
  * https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/buyer/requirements.feature#L6-L5
* Altered expected date to be inline with new date format


idavidmcdonald:
Wow, great work! 👍

PR #273, opened 2017-05-03
Skip tests on certain stages

samuelhwilliams:
## Summary
Two tests are currently failing on preview because we have some text on the homepage that changes depending on a framework state. This PR adds support for skipping specific tests based on the stage they're run against, and then duplicates (with text changes) the two failing tests and forks them to the correct branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,22 @@ To run features with a specific tag run
 
 To include or exclude tags see [the cucumber documentation](https://github.com/cucumber/cucumber/wiki/Tags#running-a-subset-of-scenarios)
 
+## Tags
+Tags are used to include/exclude given tests on certain environments. The following tags are currently supported:
+
+| Tag Name                    | Description                                           |
+|-----------------------------|-------------------------------------------------------|
+| smoke-tests                 |                                                       |
+| opportunities               |                                                       |               
+| requirements                |                                                       |
+| brief-response              |                                                       |
+| with-production-<type>-user |                                                       |
+| skip                        | Skip this test everywhere (e.g. temporarily disabled) |
+| skip-preview                | Will not run on the preview environment.              |
+| skip-staging                | Will not run on the staging environment.              |
+| skip-production             | Will not run on the production environment.           |
+
+
 ## Run tests against local services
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -47,6 +47,16 @@ To run features with a specific tag run
 
 To include or exclude tags see [the cucumber documentation](https://github.com/cucumber/cucumber/wiki/Tags#running-a-subset-of-scenarios)
 
+## Tags
```
pcraig3: David says good work on putting this into the README.

samuelhwilliams: Should've been more thorough... Will add the rest of the tags now.

idavidmcdonald: https://github.gds/gds/digitalmarketplace-jenkins/blob/26c0796acf3e3fed9611c718c1e47fe92e36a261/job_definitions/functional_tests.yml

So we currently have a generic skip tag too which is skipping on jenkins runs rather than runs from a local machine.

Paul and I kind of liked it being on jenkins rather than local because it means devs are reminded every time when running on local that we still haven't fixed some tests. This argument doesn't feel as important for environment specific skip tags.

Maybe we should change it to follow your convention and put it in the Makefile? Consistency is probably preferred. If so, we could also maybe add it to your tags readme section.

samuelhwilliams: Sounds sensible; I'll do that.

idavidmcdonald: Note, we don't _think_ we have any `skip` tags in our repository's at the moment so another approach is we just remove the `skip` tag from the jenkins jobs.

We have also spotted a `wip` tag being used in the legacy functional tests which is being excluded by jenkins at the moment
https://github.gds/gds/digitalmarketplace-jenkins/blob/26c0796acf3e3fed9611c718c1e47fe92e36a261/job_definitions/functional_tests_legacy.yml

pcraig3:
Cool, thanks for doing this! 
We probably don't need to mention the `@wip` tag at all since it's on a dead branch, so let's 🚢  this.

👍

pcraig3:
David and I have been discussing whether we should just remove the `@skip` tag completely (possibly until we need it again), but as far as this pull request is concerned, that's totally fine.

PR #274, opened 2017-05-03
Source env files if they exist

samuelhwilliams:
A local.sh file is supplied to populate environment variables so tests can easily be run locally. When the tests are run via jenkins, these environment variables are populated automatically.

This will source a local.sh if it exists and skips that step if running against preview/staging/prod.


pcraig3:
Looks good, I think. 👍

PR #277, opened 2017-05-03
Delete brief clarification scenario from legacy tests

TheDoubleK:
There is a corresponding PR against master (~~coming soon from @idavidmcdonald~~ https://github.com/alphagov/digitalmarketplace-functional-tests/pull/275) with a new version of this test, so it can be deleted from "jenkins" branch. 

Also removes two now-unused step definitions.

There are some test failures here running against preview at the moment but I'm pretty sure they're related to the recent homepage link text changes.


pcraig3:
#win 

👍

PR #284, opened 2017-05-16
Update skip tags

samuelhwilliams:
## Summary
G-Cloud 9 is now live on staging, and the functional tests are failing. This PR moves the `skip-staging` tags to the correct tests for live G9.


idavidmcdonald:
👍 

Good work. Also, I'm a fan of the skip-environment tags which I think was your suggestion. Seems to be working well!

PR #287, opened 2017-05-30
Look for a random dos 2 brief

Wynndow:
The common step to get a random dos brief was looking for a dos 1 brief.
I've just cleaned up a lot of old briefs on preview, which included
deleting all old dos 1 briefs.


TheDoubleK:
:boom: :+1:

PR #290, opened 2017-06-01
Prevent non-GET API calls for @smoke-test runs

allait:
We've had an issue with smoke test jobs accidentally including
scenarios that are creating or modifying data in the database.

Since smoke tests are supposed to be side-effect free so that we
can safely and repeatedly run them on production we need a better
way to highlight any attempts to modify data during their run.

This adds a check the the call_api method to only allow GET requests
from scenarios tagged with `@smoke-tests`. This method is used by
most data setup step definitions.

This also adds a way to override the restriction by setting the
`safe_for_smoke_tests` option when calling `call_api`. This
override is currently only used for setting up read-only users in
all environments.

Note: since this check only applies to the API calls made through
`call_api` it doesn't guarantee that the smoke test runs don't
modify any data. It's still possible to make changes using either
different methods that would send API requests or by running
steps that make modifications trough the user interface.

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -1,6 +1,10 @@
 require "rest_client"
 
 def call_api(method, path, options={})
+  safe_for_smoke_tests = options.delete(:safe_for_smoke_tests)
+  if @SMOKE_TESTS && method != :get && !safe_for_smoke_tests
+    raise "Unsafe API request in smoke tests. Only GET methods are allowed"
+  end
   domain = options.delete(:domain) || dm_api_domain
   auth_token = options.delete(:auth_token) || dm_api_access_token
   url = "#{domain}#{path}"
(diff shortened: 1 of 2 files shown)


features/support/api_helpers.rb
```diff
@@ -52,15 +56,15 @@ def ensure_user_exists(user_details)
   creation_response = call_api(:post, "/users", payload: {
     users: user_details,
     updated_by: "functional tests",
-  })
+  }, safe_for_smoke_tests: true)
```
idavidmcdonald: Is this one safe for the smoke tests? It could actually create a user with this first call right?

idavidmcdonald: Appears to only be used for the `Given /^I have a production ([a-z-]+) user` step. I didn't realise we actually had user accounts for production smoke tests... but if so then I guess it's OK? Feel's a little weird as you have to assume creating user accounts on production is always an OK thing to do as any new feature could use this step

allait: Yes. It creates 1 specific user for each role and only if that user doesn't already exist in the environment. This is an alternative to preconfiguring environment users manually before running the tests, which just made running the tests locally harder.

allait: I guess the thing to highlight is that this method is only used by `I have a production ... user` step definition and probably shouldn't be used by anything else. Most other steps just create a new random user.

idavidmcdonald:
👍

benvand:
:+1: Sorry, got sidetracked and realised I hadn't thumbsed

PR #293, opened 2017-06-12
Test 'Publish requirements' button not visible for incomplete drafts

CrystalPea:
#### Trello card:
https://trello.com/c/NvaIQwOs/539-functional-test-the-publish-button-shouldnt-be-visible-for-incomplete-requirements   
      
#### Superseded legacy test:
https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/buyer/buyer_create_requirements.feature#L27

### DIFF
--- a/features/buyer/requirements.feature
+++ b/features/buyer/requirements.feature
@@ -127,6 +127,7 @@ Scenario: Create user research participants
    Then 'Describe question and answer session' should be ticked
 
    When I click 'Review and publish your requirements'
+   Then I see the 'Publish requirements' button
     And I click 'Publish requirements'
 
   Then I don't see the 'Title' link
@@ -179,3 +180,10 @@ Scenario: Delete a draft requirement
   Then I see a destructive banner message containing 'Are you sure you want to delete these requirements?'
   When I click 'Yes, delete'
   Then I see a success banner message containing 'were deleted'
+
+
+Scenario: There is no 'Publish requirements' button for an incomplete requirement draft
+  Given I am logged in as a buyer user
+  And I have created an individual specialist requirement
+  When I click 'Review and publish your requirements'
+  Then I don't see the 'Publish requirements' button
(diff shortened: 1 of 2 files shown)


features/buyer/requirements.feature
```diff
@@ -164,3 +165,10 @@ Scenario: Delete a draft requirement
   Then I see a destructive banner message containing 'Are you sure you want to delete these requirements?'
   When I click 'Yes, delete'
   Then I see a success banner message containing 'were deleted'
+
+@no_publish_button
```
galund: The others aren't given a label in this way, is is necessary?

features/step_definitions/common_steps.rb
```diff
@@ -181,10 +181,9 @@
   breadcrumb.text().should == breadcrumb_text
 end
 
-Then /^I (don't |)see the '(.*)' link$/ do |negative, link_text|
-  page.should have_link(link_text) if negative.empty?
-
-  page.should_not have_link(link_text) unless negative.empty?
+Then /^I (don't |)see the '(.*)' (button|link)$/ do |negative, selector_text, selector_type|
+page.should have_selector(:link_or_button, selector_text) if negative.empty?
```
galund: Previously, if someone had converted a link to some other kind of element, then the test would presumably have failed.

Now, the test passes if the element is there are all, even if it isn't either a button or a link?

Seems like maybe we need a way to test based on selector_type (at which point, I'm not completely certain that doing this all in one function is the right thing, not sure :) ).

CrystalPea: It will only accept either a link or a button, no other selector, here's an overview of this logic with references:
https://stackoverflow.com/questions/13865625/capybara-matcher-for-presence-of-button-or-link

features/buyer/requirements.feature
CrystalPea: It is not necessary, I got rid of the tag now :).

features/step_definitions/common_steps.rb
galund: Ah cool, understood. Weird selector!

PR #294, opened 2017-06-12
Test that buyer can see their requirements on dashboard

CrystalPea:
Trello card:
https://trello.com/c/q9b3WiQE/531-functional-test-ensure-requirements-visible-in-buyer-dash
Superseded legacy test:
https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/buyer/buyer_create_requirements.feature#L21

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -166,7 +166,7 @@ When(/^I choose a random uppercase letter$/) do
   puts "letter: #{@letter}"
 end
 
-# the rescue is used because if a banner cannot be found the function throws an exception and does not look for the other option  
+# the rescue is used because if a banner cannot be found the function throws an exception and does not look for the other option
 Then /^I see a (success|warning|destructive|temporary-message) banner message containing '(.*)'$/ do |status, message|
   begin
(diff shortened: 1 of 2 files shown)


CrystalPea:
@Wynndow thanks a lot, this is changed now so test looks for brief in a specific summary table

features/step_definitions/common_steps.rb
```diff
@@ -240,6 +240,17 @@
   end
 end
 
+Then /^I see '(.*)' in the '(.*)' summary table$/ do |content, table_heading|
+  result_table_location = "//*[@class='summary-item-heading'][normalize-space(text())='#{table_heading}']/following-sibling::table[1]"
+  result_table_rows_location = result_table_location + "/tbody/tr[@class='summary-item-row']"
+  result_table_rows = all(:xpath, result_table_rows_location)
+  content_found = false
```
Wynndow: Nice! You could do something like the below instead of lines 247-251, but it's up to you!
  
`result_table_rows.any? {|row| row.text.include? content}.should be true`

CrystalPea: Neat suggestion, I included it :)

PR #296, opened 2017-06-14
Add new feature: suppliers being able to view their DOS services

CrystalPea:
This has been done across all DOS lots.
Trello card: https://trello.com/c/iptxy2Oy/397-allow-dos-suppliers-to-view-their-services


features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,27 @@
...
+  And I have a supplier
+  And that supplier is logged in
+  Given that supplier is on that framework
+
+Scenario Outline: Supplier coming from dashboard to view the detail page for their <lot_slug> service
+  Given that supplier has a service on the <lot_slug> lot
+  And I am on the /suppliers page
+  When I click 'View'
+  Then I am on the 'Current services' page
+  When I click '<service_name>'
```
TheDoubleK: Where do the service names come from?  I think for DOS services only the "user research studios" lot has a `serviceName` field.

Is there some magic that takes random values from the `Examples` table and inserts them in the generated services?  If so that's pretty cool.  But I still don't understand where the `service_name` lives in the data for the outcomes/specialists/participants services?

features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,27 @@
...
+
```
TheDoubleK: Probably omit this blank line if it's still the same scenario?

features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,27 @@
+@view_dos_service
```
CrystalPea: I renamed the tag so it will pertain to the who file when more tests are added

features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,27 @@
...
+Feature: Supplier being able to view their DOS services
+
+Background:
+  Given I have a live digital outcomes and specialists framework
+  And I have a supplier
+  And that supplier is logged in
+  Given that supplier is on that framework
```
CrystalPea: It was copied from a test for supplier making an application. I changed it now.

features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,27 @@
...
+
+Scenario Outline: Supplier coming from dashboard to view the detail page for their <lot_slug> service
```
CrystalPea: Scenario Outline uses Examples to parametrize and as such it's like a collection of Scenarios, it's pretty cool :). Still, I removed lot_slug from the title, as more general title is fine, you're right.

features/supplier/view_and_edit_services.feature
CrystalPea: service_names come from Examples, yes :). This is what Scenario Outline does.

features/supplier/view_and_edit_services.feature
CrystalPea: I had a blank line for legibility as I saw it used this way in requirements.feature, but I deleted it now.

features/step_definitions/common_steps.rb
```diff
@@ -256,8 +256,9 @@
   end
 end
 
-Then /^I see '(.*)' text on the page/ do |expected_text|
-  all(:xpath, "//*[normalize-space() = '#{expected_text}']").length.should >= 1
+Then /^I (don't |)see '(.*)' text on the page/ do |negative, expected_text|
+  all(:xpath, "//*[normalize-space() = '#{expected_text}']").length.should >= 1 if negative.empty?
```
CrystalPea: I changed this as per your advice, previously I had the syntax copied from another step definition with an optional negative.

features/supplier/view_and_edit_services.feature
TheDoubleK: Great - this is a cool feature, haven't seen it before.

features/support/fixtures.rb
```diff
@@ -61,6 +61,37 @@ def self.user_research_participants_service
     }
   end
 
+  def self.user_research_studios_service
+    return {
+      # `nil` values should be updated within the step when this fixture is used
+      id: nil,
```
TheDoubleK: Sorry, final thing, is there a reason to the order of keys in this example?  Otherwise it would be nice to alphabetise them, maybe.

features/supplier/view_and_edit_services.feature
```diff
@@ -0,0 +1,26 @@
...
+  And I don't see the 'Edit' link
+  And I don't see 'Remove this service' text on the page
+  And I see '<expected_content>' in the '<summary_table_name>' summary table
+
+  Examples:
+    | lot_slug                   | service_name               | summary_table_name          | expected_content |
+    | digital-specialists        | Digital specialists        | Individual specialist roles | Agile coach      |
+    | digital-outcomes           | Digital outcomes           | Team capabilities           | Agile coaching   |
+    | user-research-participants | User research participants | Recruitment approach        | Entirely offline |
+    | user-research-studios      | GDSvieux Innovation Lab    | Lab address                 | GDSbury          |
```
TheDoubleK: For the lab address the `expected_content` only has part of the address.  Is it worth including the building and postcode too?  Or will that make the example table really horrible?

features/supplier/view_and_edit_services.feature
CrystalPea: I think it's pretty cool, too. David has shown it to me :).

features/supplier/view_and_edit_services.feature
CrystalPea: It would be difficult because different parts of address are different key-value pairs and they are shown on different lines so the checker will not find them as one unbroken portion of text.

features/support/fixtures.rb
CrystalPea: I just copied them from the database and pasted in the same order they were there. Now I ordered them :)

TheDoubleK:
Sorry for the delay, didn't see this earlier.

features/support/fixtures.rb
```diff
@@ -61,6 +61,37 @@ def self.user_research_participants_service
...
+      labStreaming: "Yes \u2013 for an additional cost",
+      labTechAssistance: "Yes \u2013 for an additional cost",
+      labTimeMin: "8 hours",
+      labToilets: true,
+      labViewingArea: "No",
+      labWaitingArea: "Yes \u2013 included as standard",
+      labWiFi: true,
+      lot: 'user-research-studios',
+      serviceName: "GDSvieux Innovation Lab",
+      supplierId: nil
```
idavidmcdonald: `supplierId` is `nil` because anyone using this fixture should override it. That is why all the `nil` fields have previously been put at the top of the fixture so at a glance a developer will see all the `nil` fields. I would suggest sticking this to this pattern and moving it up.

CrystalPea: I will reorder that and then wait. I did some squishing at some point and wanted to do more but had some issues with rebasing, maybe we can look at rebasing together when this is ready for merging :).

idavidmcdonald:
Oh actually No. I've just run this on my local and the test fails... will investigate

idavidmcdonald:
Bug fixed. Have run against preview and tests now pass ✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨

PR #297, opened 2017-06-14
Functional tests for the opportunities overview page

benvand:


features/support/api_helpers.rb
```diff
@@ -210,7 +243,7 @@ def create_live_service(framework_slug, lot_slug, supplier_id, role=nil)
   random_service_id = rand(start..last).to_s
 
   service_data = {
-    updated_by: 'functional_tests',
+    updated_by: 'functiondl_tests',
```
TheDoubleK: Typo here?

features/step_definitions/brief_response_steps.rb
```diff
@@ -24,7 +24,7 @@
 end
 
 Given 'that supplier is on that framework' do
-  submit_supplier_declaration(@framework['slug'], @supplier["id"], {})
+  submit_supplier_declaration_and_sign_agreement(@framework['slug'], @supplier["id"], {'status': 'complete'})
```
TheDoubleK: I think it would be better to have a new step `And that supplier has returned a signed agreement`, which can follow on from `that supplier is on that framework` in the scenario/setup steps.

If "that supplier is on that framework" also submits the agreement then we won't easily be able to build out functional tests for the signing agreement flow - i.e. to test that flow we'll need a supplier on the framework who hasn't already submitted their agreement.

And being on a framework and having returned the thing are conceptually and prctically different things.

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -280,3 +280,47 @@ Scenario: Supplier asks a clarification question
   And I click 'Ask question'
   Then I see a success banner message containing 'Your question has been sent.'
 
+@sign-framework-agreement-call-to-action
```
TheDoubleK: I don't think we generally have individual tags per scenario?  Unless there's a reason to start doing this?

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -280,3 +280,47 @@ Scenario: Supplier asks a clarification question
...
+Scenario: Supplier can see sign framework agreement call to action
+  Given I have a live digital outcomes and specialists framework
+  And that supplier is on that framework
+  And that supplier has a service on the digital-specialists lot
+  And I have a live digital-specialists brief
+  And that supplier has filled in their application but not submitted it
+  When I click the 'View your account' link
+  Then I see the 'You must sign the framework agreement to sell these services' link
```
TheDoubleK: OK now I'm confused... does this step still pass now that `that supplier is on that framework` also submits the framework agreement?  If so, how? That message shouldn't appear if the FA has been submitted.

features/step_definitions/brief_response_steps.rb
```diff
@@ -35,6 +35,10 @@
   @brief_response = create_brief_response(@brief['lotSlug'], @brief_id, @supplier['id'])
 end
 
+Given 'that supplier submits their application' do
```
TheDoubleK: This step definition is too vaguely worded and needs clarification - "submits their application" seems like, if anything, it would be to do with applying for a framework, not applying for an opportunity.  Would be better as something like "that supplier submits their response to that brief".

The functional tests are shared by all teams so we need to be really clear what things are doing and what they apply to.  "their application" in the context of briefs team is very different from "their application" in the context of submissions.

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -280,3 +280,47 @@ Scenario: Supplier asks a clarification question
   And I click 'Ask question'
   Then I see a success banner message containing 'Your question has been sent.'
 
+@sign-framework-agreement-call-to-action
+Scenario: Supplier can see sign framework agreement call to action
+  Given I have a live digital outcomes and specialists framework
+  And that supplier is on that framework
+  And that supplier has a service on the digital-specialists lot
+  And I have a live digital-specialists brief
+  And that supplier has filled in their application but not submitted it
```
TheDoubleK: I see this step already exists merged in, but I'd **really** like to see it renamed to be clear that it is to do with an application for an opportunity, and not some other sort of application. Something like "that supplier has filled in their response to that brief but not submitted it".

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -280,3 +280,47 @@ Scenario: Supplier asks a clarification question
...
+  When I click the 'View your account' link
+  Then I see the 'You must sign the framework agreement to sell these services' link
+
+@opportunities-overview
+Scenario: Supplier can see the link to the opportunities overview
+  Given I have a live digital outcomes and specialists framework
```
TheDoubleK: There's a lot of repetition of 
```
Given I have a live digital outcomes and specialists framework
And that supplier is on that framework
And that supplier has a service on the digital-specialists lot
And I have a live digital-specialists brief
And that supplier has filled in their application but not submitted it
```

Might be worth moving it to a setup step rather than repeating in each scenario?

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -280,3 +280,47 @@ Scenario: Supplier asks a clarification question
...
+  And that supplier is on that framework
+  And that supplier has a service on the digital-specialists lot
+  And I have a live digital-specialists brief
+  And that supplier has filled in their application but not submitted it
+  When I click the 'View your account' link
+  Then I see the 'View your opportunities' link
+
+@opportunities-overview
+Scenario: Supplier can see the empty string on the opportunities overview
```
TheDoubleK: It feels like this scenario fully encompasses the previous one - i.e. if I can click the 'View your account" link then I must have been able to see it.  So I'm not sure we need both scenarios.

features/support/api_helpers.rb
```diff
@@ -95,7 +95,7 @@ def ensure_no_framework_agreements_exist(framework_slug)
 
 def update_framework_agreement_status(framework_slug, supplier_id, status)
   response = call_api(:post, "/suppliers/#{supplier_id}/frameworks/#{framework_slug}", payload: {
-    frameworkInterest: {agreementReturned: status},
+    frameworkInterest: {onFramework: status},
```
TheDoubleK: Looks like this method is no longer doing what it says?  i.e. it's called `update_framework_agreement_status` but it actually flips whether the supplier is on the framework or not, and doesn't touch the agreement status at all?

TheDoubleK: Yep, this needs to be renamed  as something like "update_supplier_on_framework".

Agreement status can no longer be set using this API endpoint (as you've discovered).

features/support/api_helpers.rb
```diff
@@ -112,14 +112,42 @@ def register_interest_in_framework(framework_slug, supplier_id)
...
+def submit_supplier_declaration_and_sign_agreement(framework_slug, supplier_id, declaration)
   path = "/suppliers/#{supplier_id}/frameworks/#{framework_slug}/declaration"
   response = call_api(:put, path, payload: {
     declaration: declaration,
     updated_by: "functional tests",
   })
   [200, 201].should include(response.code), _error(response, "Failed to submit supplier declaration #{framework_slug} #{supplier_id}")
   JSON.parse(response.body)['declaration']
+  update_framework_agreement_status(framework_slug, supplier_id, true)
+  path = "/agreements"
```
TheDoubleK: Yep, this is the "new" (although in reality quite old) way we need to update agreement status.  It should definitely be in it's own step definition and not bundled in here.

features/support/api_helpers.rb
```diff
@@ -112,14 +112,42 @@ def register_interest_in_framework(framework_slug, supplier_id)
...
-def submit_supplier_declaration(framework_slug, supplier_id, declaration)
+def submit_supplier_declaration_and_sign_agreement(framework_slug, supplier_id, declaration)
   path = "/suppliers/#{supplier_id}/frameworks/#{framework_slug}/declaration"
   response = call_api(:put, path, payload: {
     declaration: declaration,
     updated_by: "functional tests",
   })
   [200, 201].should include(response.code), _error(response, "Failed to submit supplier declaration #{framework_slug} #{supplier_id}")
   JSON.parse(response.body)['declaration']
+  update_framework_agreement_status(framework_slug, supplier_id, true)
```
TheDoubleK: This "update_supplier_on_framework" (as it actually now is) also feels like a separate step, something like "And the supplier has been awarded onto the framework", which is a thing that happens after they submit their declaration.

I'm not sure it's 100% necessary to set the `on_framework` flag at all though, because I don't think our frontend apps use it - it's mainly used in scripts.  But I could be wrong about that.

features/support/api_helpers.rb
benvand: Fixed

features/support/api_helpers.rb
benvand: Renamed to `set_supplier_on_framework`

features/support/api_helpers.rb
benvand: Supplier frontend uses it to decide whether to show the 'View documents' and 'View your opportunities' links. Will do the renaming shortly.

features/supplier/supplier_applies_for_an_opportunity.feature
benvand: It didn't but it does now ive separated them

features/supplier/supplier_applies_for_an_opportunity.feature
benvand: I quite like it because you can run a single test without editing the file but I'm not married to it

benvand:
@TheDoubleK Hi Kev I've made the requested changes if you fancy having another look

features/support/api_helpers.rb
```diff
@@ -120,6 +120,37 @@ def submit_supplier_declaration(framework_slug, supplier_id, declaration)
   })
   [200, 201].should include(response.code), _error(response, "Failed to submit supplier declaration #{framework_slug} #{supplier_id}")
   JSON.parse(response.body)['declaration']
+  set_supplier_on_framework(framework_slug, supplier_id, true)
```
TheDoubleK: I don't think setting `on_framework=True` should be part of `submit_supplier_declaration` - you can submit a declaration and then not be awarded onto a framework.

I think it makes more sense to the `that supplier is on that framework` step - it could submit the declaration (as it already does) and then have an additional call to `set_supplier_on_framework`.

features/step_definitions/brief_response_steps.rb
```diff
@@ -23,18 +23,30 @@
   page.visit("#{dm_frontend_domain}#{url}")
 end
 
-Given 'that supplier is on that framework' do
-  submit_supplier_declaration(@framework['slug'], @supplier["id"], {})
+Given 'that supplier has applied to be on that framework' do
+  submit_supplier_declaration(@framework['slug'], @supplier["id"], {'status': 'complete'})
+end
+
+Given 'we accept that suppliers application to the framework' do
```
TheDoubleK: These two should probably be `And`s rather than `Given`s, but we don't really have a system for this in the step definitions so let's not worry about it right now.

TheDoubleK:
Cheers Ben, sorry this turned into a bit of a marathon but I think it's good for it, this is much clearer now I think :+1:

PR #298, opened 2017-06-19
Add tests for creating a draft requirement for all lots

CrystalPea:
Trello card: https://trello.com/c/cAcMOYus/708-functional-test-buyer-creating-a-draft-requirement

Superseded legacy test: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/buyer/buyer_create_requirements.feature#L8

### DIFF
--- a/features/buyer/requirements.feature
+++ b/features/buyer/requirements.feature
@@ -4,6 +4,25 @@ Feature: Create and publish a requirement
   As a buyer within government
   I want to be able to create and publish a requirement
 
+Scenario Outline: Create a draft requirement for each lot
+  Given I am logged in as a buyer user
+  When I click '<link_name>'
+  Then I am on the '<link_name>' page
+  When I click 'Create requirement'
+  Then I am on the 'What you want to call your requirements' page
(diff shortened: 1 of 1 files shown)


alext:
@CrystalPea I don't think you meant me...

CrystalPea:
@alext you're totally right, it was a misclick and I have rectified it after a few seconds but I guess Github has sent an email to you already 😅

CrystalPea:
@idavidmcdonald 
Thanks for commenting. I thought we discussed this before writing this test and then the conclusion was that `And I have created an individual specialist requirement` is a bit different since it creates the draft by direct API call instead of clicking through from the homepage. This test checks if links on homepage work and if creating the draft by clicking links and filling in the fields works. Is this checked elsewhere?

idavidmcdonald:
Sorry, that was my assumption based on the step name but I was wrong. When I looked a bit further just now it turns out it does go through the form

https://github.com/alphagov/digitalmarketplace-functional-tests/blob/835df382a031b4c27b089cdd418a9c079e00a5b8/features/step_definitions/requirements_steps.rb#L5

PR #300, opened 2017-06-21
G9 tests

galund:
Tests to replace those lost in G9 (e.g. see <https://github.com/alphagov/digitalmarketplace-functional-tests/pull/267>).

One of the tests fails because of a bug, so maybe we shouldn't land this yet ;)

We are just picking random filters and random categories, so these are not exhaustive tests like those they replace. But we think that known-input known-output exhaustive testing really belongs as unit tests in the buyer frontend rather than here.

### DIFF
--- a/features/smoke-tests/buyer/catalogue.feature
+++ b/features/smoke-tests/buyer/catalogue.feature
@@ -71,3 +71,27 @@ Scenario: User is able to click on a random category
   And I see that category_name in the search summary text
   And I see a search result
   And I see fewer search results than before
+
+Scenario: User is able to click on several random filters
+  Given I am on the /g-cloud page
+  And I have a random g-cloud lot from the API
+  And I click that lot.name
+  Then I am on the 'Search results' page
+  And I note the number of search results
+  Then I select several random filters
+  And I click 'Filter'
+  Then I am on the 'Search results' page
+  And I see fewer search results than before
+
+Scenario: User is able to paginate through search results and all of the navigation is preserved
+  Given I am on the /g-cloud page
+  And I have a random g-cloud lot from the API
+  And I click that lot.name
+  Then I am on the 'Search results' page
+  And I note the number of category links
+  And I click 'Next'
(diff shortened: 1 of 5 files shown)


features/smoke-tests/buyer/catalogue.feature
```diff
@@ -58,3 +58,40 @@ Scenario: User is able to search by keywords field on the search results page to
...
+Scenario: User is able to click on a random category
+  Given I am on the /g-cloud page
+  And I have a random g-cloud lot from the API
+  And I click that lot.name
+  Then I am on the 'Search results' page
+  And I note the number of search results
+
+  When I click a random category link
+  Then I am on the 'Search results' page
+  And I see that category_name in the search summary text
```
samuelhwilliams: It makes me a little sad that we have `lot.name` and `category_name` in the same block.

features/smoke-tests/buyer/catalogue.feature
```diff
@@ -58,3 +58,40 @@ Scenario: User is able to search by keywords field on the search results page to
...
+
+Scenario: User is able to paginate through search results and all of the navigation is preserved
+  Given I am on the /g-cloud page
+  And I have a random g-cloud lot from the API
+  And I click that lot.name
+  Then I am on the 'Search results' page
+  And I note the number of category links
+  And I click 'Next'
+  Then I am taken to page 2 of results
+  And I see the same number of category links as before
```
samuelhwilliams: Maybe `as noted` would be clearer; chaining multiple `before`s makes it less clear which value is being compared against.

features/smoke-tests/buyer/catalogue.feature
galund: Yeah I think this is justifiable: the "I click a random category link" step literally just scrapes the category name off the page, so we don't have anything approaching an object that properly represents a category. (We could put it into a dictionary, with probably just the one key, but I think that would set wrong expectations, so is less clear.)

PR #301, opened 2017-07-05
Add default.nix

risicle:
See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/719 for general discussion.

In this repo we take a slightly different approach. I've use a nix tool called `bundix` to generate a nix file (`gemset.nix`) from the `Gemfile` and `Gemfile.lock`. This means that the gems get installed as actual nix packages, inheriting some of the benefits of doing so (notably not *all*). `bundix` is included in this env to make it possible for people to regenerate `gemset.nix`.

Disadvantages of this approach are that the nix file needs to be kept in sync with the `Gemfile.lock` to remain useful and it's not always obvious when things have slipped slightly out of sync. Also you're no longer using the same tool to install your language-level dependencies everywhere so it may not be obvious when subtle problems creep in which don't affect the nix environment, but do affect a direct-bundler environment.

If we really loved this approach, it *is* possible to do the same thing for python projects `requirements.txt` files using a tool called `pypi2nix` (I'm not super-keen on it personally).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 reports/
 config/
 sql/data/
+local.nix
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,24 @@
+{
+  pkgs ? import <nixpkgs> {}
+}:
+{
+  digitalMarketplaceFunctionalTestsEnv = (pkgs.bundlerEnv {
+      name = "digitalmarketplace-functional-tests-bundler-env";
+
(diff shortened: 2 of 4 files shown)


risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

TheDoubleK:
:+1:

PR #302, opened 2017-07-10
Fix test fail when there is a service with 'Next' in the name.

galund:
Being explicit with which links we want to click on, because of some services having the word 'next' in their name (!)

https://trello.com/c/5t96gz0Z/400-functional-tests


galund:
I'm not convinced because I don't think even in this case the link says
exactly "Next" (it also includes the page number and count?) Happy to try
though and see how many existing tests break :) ..

samuelhwilliams:
Ah duh yeah. Ok. Don't know why you've used `klass` rather than - I assume - `class` - but if this works then looks good.

PR #304, opened 2017-07-11
Add skip environment tags for link text that is being deployed

idavidmcdonald:
This is the supplier frontend PR that broke the tests - https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/714

### DIFF
--- a/features/supplier/view_and_edit_services.feature
+++ b/features/supplier/view_and_edit_services.feature
@@ -7,10 +7,30 @@ Background:
   And that supplier is logged in
   And that supplier is on that framework
 
+@skip-staging
 Scenario Outline: Supplier coming from dashboard to view the detail page for one of their services
   Given that supplier has a service on the <lot_slug> lot
   And I am on the /suppliers page
   When I click 'View services'
+  Then I am on the 'Current services' page
+  When I click '<service_name>'
+  Then I am on the '<service_name>' page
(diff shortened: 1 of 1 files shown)


benvand:
Should the skip tags be the other way around?
The older, existing functionality will be on staging, and should therefore be skipped on preview.
The newer test, with the functionality we're introducing, will be available on preview only and so we need to skip it on staging.

idavidmcdonald:
I think it is correct as it stands?

At the moment on preview we have the text 'View services'
At the moment on staging we have the text 'View'

MuriloDalRi:
The test looks good 👍

features/supplier/view_and_edit_services.feature
```diff
@@ -24,3 +25,22 @@ Scenario Outline: Supplier coming from dashboard to view the detail page for one
     | digital-outcomes           | Digital outcomes           | Team capabilities           | Agile coaching   |
     | user-research-participants | User research participants | Recruitment approach        | Entirely offline |
     | user-research-studios      | GDSvieux Innovation Lab    | Lab address                 | GDSbury          |
+
+@skip-preview
```
samuelhwilliams: Be nice to throw in `@skip-local` as well, for anyone who tries to run the functional tests locally.

PR #305, opened 2017-07-13
Docker compose

Wynndow:
Add a Docker Compose configuration for helping with setting up a local environment for running functional tests again. See the README for more info.


.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+
+!local_nginx/
```
allait: There's already an `nginx/` folder, can we use that?

README.md
```diff
@@ -78,3 +79,52 @@ applications are running the functional tests can be run with
...
+## Setting up your local environment with Docker Compose to run functional tests against
+
+Setting up your local environment and database to run the functional tests against can be a pain. Wouldn't it be nice if you could get all the apps up and running, backed by a database in the correct state with just one command? That's hopefully what we can do with Docker Compose. It uses the latest tagged Docker images of the apps as well as a DB dump of your choosing to achieve this.
+
+### Setup
+
+  - Have Docker installed on your machine - see [here](https://www.docker.com/docker-mac)
+  - Have Docker Compose installed on your machine. Docker for Mac comes with it so you probably won't need to do anything extra. You can check with `docker-compose -v`. If you don't have it, go [here](https://docs.docker.com/compose/install/)
+  - Have a database dump that you want to use. The most up to date you can get your hands on the better. This is the data you're going to be running tests against so it should have all frameworks up to date.
+  - In the root of the functional tests repo create a directory called sql and place your database dump in there. Make sure the dump has an `.sql` suffix.
```
Wynndow: The solution for the db dump thing means that there is now an `sql` dir. I've added the name of the dump to `.gitignore` though.

README.md
```diff
@@ -78,3 +79,52 @@ applications are running the functional tests can be run with
...
+  - Execute `$ make docker-up`
```
Wynndow: Cool

README.md
```diff
@@ -78,3 +79,52 @@ applications are running the functional tests can be run with
...
+  - Enter your AWS MFA code when prompted.
+  - If this is the first time you've run this command, the Postgres image will need to import the data from the dump. This will take a minute or two. When it's done, the apps will load, beginning with the api. When the logs in your terminal have calmed down and tell you that frontend apps have spawned uWSGI workers, move on.
+  - Go to [`http://localhost`](http://localhost). You should see the Digital Marketplace homepage.
+  - You now need to index your services. You can do this with a script in the scripts repo. From the root of your scripts repo, you're probably going to want to do this:
+        ./scripts/index-services.py  dev --api-token=myToken --search-api-token=myToken  --frameworks=g-cloud-9
```
Wynndow: It'd have to work on everyones system though. If there scripts repo is on a different relative path it'd break, right?

README.md
```diff
@@ -78,3 +79,52 @@ applications are running the functional tests can be run with
...
+### Updating the images
+      $ docker-compose pull
+  - Docker Compose will use app images tagged with `:latest`.
+  - If it finds them locally it'll just use them.
+  - Before running tests you'll probably want to pull the latest versions of the apps.
+  - This command will pull them all.
+
+### Running tests against YOUR version of the app.
+  - The main reason for al this is to make it easier to test YOUR local changes, which won't happen if you always have the `:latest` tagged image.
+  - Building your app locally will create a local image of it and give it the `:local` tag.
```
Wynndow: Yeah, typo. Should have been `:latest`

README.md
```diff
@@ -78,3 +79,52 @@ applications are running the functional tests can be run with
...
+  - Docker Compose will now use your local copy including your new stuff.
+  - From the root of the app you want to build an image for, run `$ make docker-build`
```
Wynndow: Done

.dockerignore
Wynndow: Yep!

Wynndow:
@allait Have added a solution for the database import errors issue, let me know what you think!

README.md
allait: Yeah, we can do the same thing we do for credentials with `DM_SCRIPTS_PATH`. It feels like setting one variable should be easier than getting all the parameters right. It's also more likely to be updated when the framework iteration changes.

Makefile
```diff
@@ -22,4 +22,10 @@ config/local.sh:
 clean:
 	rm -rf reports/
 
-.PHONY: smoke-tests run rerun setup install clean
+docker-up:
+	$(eval export AWS_ACCESS_KEY_ID=$(shell aws configure get aws_access_key_id))
+	$(eval export AWS_SECRET_ACCESS_KEY=$(shell aws configure get aws_secret_access_key))
+	$(eval export DM_MANDRILL_API_KEY=$(shell ${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/vars/preview.yaml | grep mandrill_key | sed 's/^.* //'))
```
allait: Make should probably stop if this fails. It prints an error that quickly gets lost in the compose output.

README.md
```diff
@@ -78,3 +79,55 @@ applications are running the functional tests can be run with
...
+
+### Setup
+
+  - Have Docker installed on your machine - see [here](https://www.docker.com/docker-mac)
+  - Have Docker Compose installed on your machine. Docker for Mac comes with it so you probably won't need to do anything extra. You can check with `docker-compose -v`. If you don't have it, go [here](https://docs.docker.com/compose/install/)
+  - Have a `.envrc` file in the root of your credentials repo which sets your AWS profile to your sops profile. Probably something like `AWS_PROFILE=sops`. Check in `~/.aws/config` if you're not sure. This step isn't strictly necessary to get all the containers up and running, but functional tests will fail whenever trying to send an email if you don't do it.
+  - Have a database dump that you want to use. The most up to date you can get your hands on the better. This is the data you're going to be running tests against so it should have all frameworks up to date.
+  - Place your database dump into the `./sql` directory.
+  - Rename your dump to `database_dump`. This needs to be exact - no suffix!
```
allait: Is that because postgres will pick up any sql files in `/docker-entrypoint-initdb.d`? I think if you map sql folder to a different directory you could modify the script to run the file from there without requiring a name. It's also very likely that it would be called `exportdata.sql` since that's what our DB export files are names.

README.md
Wynndow: Yeah, that makes sense. I'll update and add a step to setup to set that env variable.

Makefile
Wynndow: To get the containers up and running it's not strictly necessary to set any of these. You just won't be able to upload documents or send fake emails. So the func tests will fail. I have a feeling that some people won't mind if this fails, they just want to play with the apps. I guess they could just use `docker-compose up` in that case.

README.md
Wynndow: Yeah exactly. It also picks up .sh files which is why I need to map the sql dir to there. I guess I could map another volume so the .sql file can sit somewhere else, and the .sh script can remain where it is?

PR #308, opened 2017-07-17
Remove skip staging tag for features now live

idavidmcdonald:
The view your opportunities feature is now on staging so we can also run the tests on that environment.


MuriloDalRi:
👍

PR #312, opened 2017-07-21
Update xpath to only find the next table

benvand:
result_table_location = "//*[@class='summary-item-heading'][normalize-space(text())=\"#{table_heading}\"]/following-sibling::table[1]

will find the next sibling table on the page. Because we don't render a table element for empty tables this means that it will find the next 'non-empty table'

For example if out buyers homepage looks like this:

```
Unpublished requirements
You don’t have any unpublished requirements

Published requirements
You don’t have any published requirements

Closed requirements
Name	Closed	Responses
Some requirement	Wednesday 15 March 2017	View responses
```

And we look for the 'Published requirements' table we will find the next sibling table of the Published requirements header.
The next sibling _table_ is actually the closed requirements table.

The new xpath finds the next sibling _anything_

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -251,7 +251,7 @@ Then /^I see the '(.*)' summary table filled with:$/ do |table_heading, table|
 end
 
 Then /^I see '(.*)' in the '(.*)' summary table$/ do |content, table_heading|
-  result_table_location = "//*[@class='summary-item-heading'][normalize-space(text())=\"#{table_heading}\"]/following-sibling::table[1]"
+  result_table_location = "//*[@class='summary-item-heading'][normalize-space(text())=\"#{table_heading}\"]/following-sibling::*[1]"
   result_table_rows_location = result_table_location + "/tbody/tr[@class='summary-item-row']"
   result_table_rows = all(:xpath, result_table_rows_location)
   result_table_rows.any? {|row| row.text.include? content}.should be true


idavidmcdonald:
Wondering if we want this step to be so flexible (the step will read like we are looking for content in a summary table when actually we are finding content in a string where a summary table usually exists). 

Should we instead be using a different step if we are not expecting to find a table that checks the string?

Not sure what the structure of our tables are but if it is something semantically like
<table wrapper div>
    <table title>
    <summary table>
</table wrapper div>
then our original table summary step could look specifically for summary tables found in our table wrapper meaning we wouldn't risk looking at a table related to a different header. If there is no wrapper for table titles and the actual table then this wouldn't be possible though.

I'd let this PR go in but would be worth having a think about the above first

benvand:
There is no table wrapper 😭 

So all we can do is find the next sibling.
We could raise if that isn't a table :man_shrugging:
I definitely thing these steps around tables could use work. It would be good to be able to check that the empty string is correct for example, which I don't think we have any function for at the moment.

benvand:
I ended up writing this in the closed brief branch because we need to check for one or more entries but can't know what they are:
https://github.com/alphagov/digitalmarketplace-functional-tests/compare/closed-brief?expand=1#diff-1bc6a645fdebd83767222491e911ee97R260

We should have a think about what we're going to need going forward and come to a consensus about what we want to do re looking stuff up in table and probably more generally.

Settle on a shared language/ way of putting this stuff together. There must be a style guide somewhere we can use.

PR #315, opened 2017-08-03
Use supplier user id for signing framework agreement

Wynndow:
### Use supplier_user_id when signing framework

The api helper to sign a framework agreement was using a hard coded ID
of `1` for the `uploaderUserId`. This is fine if there is a user with an ID
of `1`. If not it fails to sign the framework agreement and the functional
test fail at a later point.

This occurred when moving cleaned up data from prod to staging.

Passing in the supplier user id an argument to the helper solves this
issue.

### Remove quotes from steps that prevents them running

There were quotes around some steps in the `view_and_edit_services`
scenarios. This means they weren't matching with the step definitions
and so weren't being run, just skipped. This is different to a failure
and so may have been missed.

Tests seem to pass with them being run so I'm correcting here.

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -32,7 +32,7 @@ Given 'we accept that suppliers application to the framework' do
 end
 
 Given 'that supplier returns a signed framework agreement for the framework' do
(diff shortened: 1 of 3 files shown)


benvand:
LGTM 👍

PR #318, opened 2017-08-08
supplier services page: temporarily disable check for supplier services listing page title

risicle:
We will need to be able to check for variable-composed headings in future which is not something we have figured out how to do yet. Suggestions welcome, but in the meantime I suggest someone merges this.


risicle:
Nice one, ben.

PR #319, opened 2017-08-09
Legacy functional tests - Supplier A-Z

DilwoarH:
We need to move legacy functional tests relating to the Supplier A-Z into the func tests master branch.

They're here: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/buyer/buyer_journey.feature

A/C: we have tests for pagination (or lack of pagination as appropriate) in the Supplier A-Z page.

Ticket: https://trello.com/c/rbNn8nps/582-legacy-functional-tests-supplier-a-z

### DIFF
--- a/features/step_definitions/supplieraz_steps.rb
+++ b/features/step_definitions/supplieraz_steps.rb
@@ -41,3 +41,41 @@ Then (/^I see that supplier in one of the pages that follow from clicking '(.*)'
     ".//h2[@class='search-result-title']/a"
   ).text.should == normalize_whitespace(@supplier['name'])
 end
+
+Given(/I navigate to the list of '(.*)' page$/) do |value|
+  page.visit("#{dm_frontend_domain}/g-cloud/suppliers?prefix=#{value.split(' ').last}")
+  @data_store = @data_store || Hash.new
+  @data_store['supplier_alphabet'] = "#{value.split(' ').last}"
+end
+
+Then(/I am taken to page '(.*)' of results$/) do |page_number|
+  if current_url.include?('suppliers?')
+    current_url.should include("#{dm_frontend_domain}/g-cloud/suppliers?")
+    current_url.should include("prefix=#{@data_store['supplier_alphabet']}")
+    current_url.should include("page=#{page_number}")
+    current_url.should include("framework=g-cloud")
+    if page_number >= '2'
+      page.should have_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+    elsif page_number < '2'
+      page.should have_selector(:xpath, "//a[contains(text(), 'Next')]//following-sibling::span[contains(text(),'page')]")
+      page.should have_no_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+    end
+  end
+end
+
+
+Then(/pagination is '(.*)'$/) do |availability|
+  if current_url.include?('suppliers?')
+    pagination_available = "#{availability}"
+  else
+    fail("Pagination check only works for the suppliers list page these days.")
+  end
+
+  if pagination_available == 'available'
+    page.should have_selector(:xpath, "//a[contains(text(), 'Next')]//following-sibling::span[contains(text(),'page')]")
+    page.should have_no_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+  elsif pagination_available == 'not available'
+    page.should have_no_selector(:xpath, "//a[contains(text(), 'Next')]//following-sibling::span[contains(text(),'page')]")
+    page.should have_no_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+  end
+end
--- a/features/smoke-tests/supplier/supplieraz.feature
+++ b/features/smoke-tests/supplier/supplieraz.feature
@@ -24,3 +24,14 @@ Scenario: User can navigate to a specific supplier's detail page
   Then I see that supplier in one of the pages that follow from clicking 'Next page'
   When I click that specific supplier
   Then I am on that supplier.name page
+
+Scenario: There is pagination on the list of suppliers page if there are more than 100 results
+  Given I navigate to the list of 'Suppliers starting with S' page
+  When I click the 'Next page' link
+  Then I am taken to page '2' of results
+  When I click the 'Previous page' link
+  Then I am taken to page '1' of results
+
+Scenario: There is no pagination on the list of suppliers page if there is less than or equal to 100 results
+  Given I navigate to the list of 'Suppliers starting with Q' page
+  Then pagination is 'not available'


features/buyer/buyer_journey.feature
```diff
@@ -0,0 +1,26 @@
+@not-production @functional-test
+Feature: Buyer user journey through Digital Marketplace
+
+Scenario: Check the admin frontend /_status
+  Given I am logged in as a admin user
```
CrystalPea: would "I am logged in as **an** admin user" also work? It would be more grammatical this way.

features/buyer/buyer_journey.feature
```diff
@@ -0,0 +1,26 @@
+@not-production @functional-test
+Feature: Buyer user journey through Digital Marketplace
```
CrystalPea: It could be useful to write that it's Buyer user's journey through GCloud, as on DOS it would be very different. I would also recommend including this in file-names, because now it's a bit confusing.

features/buyer/buyer_journey.feature
```diff
@@ -0,0 +1,26 @@
...
+  Given I am logged in as a admin user
+  Then I am on the frontend /admin/_status page
+  Then I see 'ok' as the value of the 'status' JSON field
+  And Display the value of the 'version' JSON field as 'Release version'
+
+Scenario: Setup for tests
+  Given I have test suppliers
+  And The test suppliers have live services
+  And All services for the test suppliers are Public
+  Then I take a screenshot
```
CrystalPea: Screenshots are used during development and then taken out usually. Unless there a reason why this one is here?
Also I don't really understand this scenario, what does this flow do? Should it be a separate scenario? Maybe what you need is "Background" feature, like here? https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/supplier/supplier_applies_for_an_opportunity.feature

features/step_definitions/buyer_journey.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+  end
+end
+
+
+Then(/Pagination is '(.*)'$/) do |availability|
+  if current_url.include?('suppliers?')
+    pagination_available = "#{availability}"
+  else
+    fail("Pagination check only works for the suppliers list page these days.")
```
Wynndow: I'm not sure this is true. We have pagination on [the dos opportunities page for example](https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities). Maybe this step could be moved to a common place and then used elsewhere.

features/buyer/buyer_journey.feature
```diff
@@ -0,0 +1,26 @@
...
+Scenario: There is pagination on the list of suppliers page if there are more than 100 results
+  Given I navigate to the list of 'Suppliers starting with S' page
+  When I click the 'Next page' link
+  Then I am taken to page '2' of results
+
+  When I click the 'Previous page' link
+  Then I am taken to page '1' of results
+
+Scenario: There is no pagination on the list of suppliers page if there is less than or equal to 100 results
+  Given I navigate to the list of 'Suppliers starting with Q' page
```
Wynndow: This feels a little fragile. We have 51 suppliers on g-cloud-9 who's name starts with Q. I guess it's probably fine for quite some time but there may be a better way to do it.

features/step_definitions/supplier_utils.rb
```diff
@@ -0,0 +1,138 @@
+#Utils
+
+# suppliers
+def create_supplier (supplier_id, supplier_name, supplier_description, supplier_contactName, supplier_email, supplier_postcode, supplier_dunsnumber)
```
Wynndow: We have [a function in our api helpers](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/support/api_helpers.rb#L220) to create suppliers. At the moment it's not paramaterized, so I think you can only create the same supplier. You may be able to update it so you can create different suppliers.

features/step_definitions/supplier_utils.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    contact_id = JSON.parse(response.body)["suppliers"]["contactInformation"][0]["id"]
+    contact_data = JSON.parse("{\"updated_by\":\"Functional tests\", \"contactInformation\":#{supplier_data ["suppliers"]["contactInformation"][0].to_json}}")
+    response = call_api(:post, "/suppliers/#{supplier_id}/contact-information/#{contact_id}", payload: contact_data, safe_for_smoke_tests: true)
+    response.code.should be(200), response.body
+  end
+end
+
+# services
+def create_live_service (supplier_id, service_id, framework, lot)
```
Wynndow: We have [this helper](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/support/api_helpers.rb#L220) already for creating services which you may be able to extend to allow creation of g-cloud services (looks like it's only doing DOS at the moment.)

features/step_definitions/supplier_utils.rb
```diff
@@ -0,0 +1,138 @@
...
+    response = call_api(:post, service_path, payload: service_data)
+    response.code.should be(200), response.body
+
+    service_data.delete("services")
+    response = call_api(:post, service_path + "/status/published", payload: service_data)
+    response.code.should be(200), response.body
+  end
+end
+
+def service_status_public (service_id)
```
Wynndow: This feels like is should be a cucumber step itself instead of just a function?

features/step_definitions/supplier_utils.rb
```diff
@@ -0,0 +1,138 @@
...
+  when "CCS Sourcing", "CCS Category", "Administrator"
+    return "Administrator"
+  when "Supplier", "Buyer"
+    return "Digital Marketplace"
+  else
+    fail("User type \"#{user_type}\" does not exist")
+  end
+end
+
+Given(/I have logged in to Digital Marketplace as a '(.*)' user$/) do |user_type|
```
Wynndow: We have [login steps already](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/step_definitions/login_steps.rb), which you should try and use instead of creating new ones.

fixtures/g-cloud-6-iaas-service.json
```diff
@@ -0,0 +1,356 @@
+{
```
Wynndow: As mentioned above, these lots are redundant so we may need new fixtures, or another way of creating g-cloud services.

fixtures/test-supplier.json
```diff
@@ -0,0 +1,27 @@
+{
```
Wynndow: As mentioned above, possible don't needs this with the current supplier creation function.

DilwoarH:
@galund @Wynndow @CrystalPea - all tests have been updated so that there are no database interactions.

features/step_definitions/supplieraz_steps.rb
```diff
@@ -41,3 +41,54 @@
...
+  @data_store = @data_store || Hash.new
+  @data_store['supplier_alphabet'] = "#{value.split(' ').last}"
+end
+
+Then(/I am taken to page '(.*)' of results$/) do |page_number|
+  if current_url.include?('search?')
+    current_url.should include("#{dm_frontend_domain}/g-cloud/search?")
+    current_url.should include("page=#{page_number}")
+    current_url.should include("lot=iaas")
+    if page_number >= '2'
```
galund: The detail of this test _only_ works if you're on page 2, in fact (it checks that the 'next' page is page 3, I think?)

I think anyway it might be worth separating this into separate tests, such as "Then I can see a Next link to page 3" etc?

features/support/env.rb
```diff
@@ -64,5 +64,13 @@ def dm_pagination_limit()
   (ENV['DM_PAGINATION_LIMIT'] || 100).to_i
 end
 
+def dm_admin_email()
```
Wynndow: I can't see where these two new helper functions are being used?

features/step_definitions/supplieraz_steps.rb
```diff
@@ -41,3 +41,54 @@
...
+  page.visit("#{dm_frontend_domain}/g-cloud/suppliers?prefix=#{value.split(' ').last}")
+  @data_store = @data_store || Hash.new
+  @data_store['supplier_alphabet'] = "#{value.split(' ').last}"
+end
+
+Then(/I am taken to page '(.*)' of results$/) do |page_number|
+  if current_url.include?('search?')
+    current_url.should include("#{dm_frontend_domain}/g-cloud/search?")
+    current_url.should include("page=#{page_number}")
+    current_url.should include("lot=iaas")
```
Wynndow: I don't think checking the lot here is applicable? We do have `framework=g-cloud` in the url now though.

features/step_definitions/supplieraz_steps.rb
DilwoarH: What if there is no page 3? the page number from the url is enough right?

DilwoarH:
Removed any code which was not required by the tests.

DilwoarH:
@Wynndow - could you let me know if this is fine now?

DilwoarH:
Hi @CrystalPea - Could you review the changes please :)

features/step_definitions/supplieraz_steps.rb
```diff
@@ -41,3 +41,41 @@
...
+      page.should have_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+    elsif page_number < '2'
+      page.should have_selector(:xpath, "//a[contains(text(), 'Next')]//following-sibling::span[contains(text(),'page')]")
+      page.should have_no_selector(:xpath, "//a[contains(text(), 'Previous')]//following-sibling::span[contains(text(),'page')]")
+    end
+  end
+end
+
+
+Then(/Pagination is '(.*)'$/) do |availability|
```
samuelhwilliams: `Pagination` should be lowercase.

PR #321, opened 2017-08-16
Add routing for /buyers/direct-award and high-level test

samuelhwilliams:
## Summary
Updates dev bootstrap script to add routing for `/buyers/direct-award` to the buyer-frontend app.

## Corresponding PR
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/581

## Ticket
https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/features/smoke-tests/buyer/direct-award.feature
+++ b/features/smoke-tests/buyer/direct-award.feature
@@ -1,0 +1,10 @@
+@skip-staging @skip-production
+@smoke-tests
+@direct-award-passive
+Feature: Direct Award passive assurance
+
+Scenario: User is required to login to save a search
+  Given I am on the /g-cloud/search page
+  Then I see the 'Save your search' button
+  And I click 'Save your search'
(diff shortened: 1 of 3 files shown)


features/smoke-tests/buyer/direct-award.feature
```diff
@@ -0,0 +1,10 @@
+@skip-staging @skip-production
+@smoke-tests
+@direct-award-passive
+Feature: Direct Award passive assurance
+
+Scenario: User can save a search into a new Direct Award Project
```
galund: I think I'd call this "User is invited to authenticate in order to safe a Direct Award search" to distinguish it from the logged-in version? (or do we always get really context of the 'feature' title, in which case maybe just clarifying that would be enough - "passive assurance" isn't so clear to me).

samuelhwilliams: Good idea to re-word it. Have done so.

PR #323, opened 2017-08-16
Include brief responses frontend

Wynndow:
This is a based on this [https://github.com/alphagov/digitalmarketplace-functional-tests/pull/322](https://github.com/alphagov/digitalmarketplace-functional-tests/pull/322). Once that's merged the stuff about finding radio buttons will disappear.

### Create brief-responses-frontend service
We add the new brief responses frontend as a service in the
docker-compose file, following the same pattern as the other frontend
apps.

### Add a new location to the nginx config
The brief-responses frontend app has a new prefix so traffic is routed
correctly.


Wynndow:
Yep, i'll stick it in.

TheDoubleK:
I have no idea what's going on with the functional tests these days.  What do they have to do with Docker?  I'm lost.

TheDoubleK:
Why are there empty "description" fields in the table of tags in the README?  Feels like they would be really really useful. (Not for this PR, I know)

Wynndow:
@TheDoubleK We've got a docker-compose file in the functional tests repo. Allows you to spin up all the apps locally easily without faff, so you can run local functional tests easily. It's all documented in the readme. Has no impact on the tests we run on Jenkins though.

TheDoubleK:
Is the docker thing only relevant for running functional tests?  Or could it be an everyday part of dev setup for running all the things?  In which case having it squirrelled away here might not be the best place for it (you can say the same thing about the nginx bootstrap script here, we really should have a better home for that).

Wynndow:
It could be useful for everyday, although I suspect it might be more of a pain than what people are currently doing. To get any changes up and running you'd need to rebuild the docker image and then restart the whole thing. Also, the logging is weird in them cos the apps are running behind supervisor/uwsgi/nginx. And using a debugger for example would be pretty difficult. I guess if people start using the compose thing we can see how they like to use it and go from there?

PR #327, opened 2017-08-18
Adapt legacy tests to handle reorganization of supplier dashboard

risicle:
Though I realize the general policy is to allow legacy functional tests to break and disable them, doing so with these changes would have probably caused quite a few downstream tests to break. So for instance one thing I did was fix the "user can login" test to check for the correct email address...


TheDoubleK:
Good stuff :+1:

PR #333, opened 2017-08-30
Ensure correct brief is selected in award test

benvand:
https://www.foreach.be/blog/9-tips-improving-cucumber-test-readability
Changed some of the `'when'`, `'then'` etc

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -32,6 +32,7 @@ Given /^I am logged in as the buyer of a closed brief with responses$/ do
   closed_brief = get_brief(submitted_brief_response['brief']['id'])
   @brief_id = closed_brief['id']
   @lot_slug = closed_brief['lotSlug']
+  @framework_slug = closed_brief['frameworkSlug']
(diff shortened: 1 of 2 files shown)


features/step_definitions/brief_response_steps.rb
```diff
@@ -44,6 +45,11 @@
   page.visit("#{dm_frontend_domain}#{url}")
 end
 
+Given /^I click the 'Tell us who won this contract' link for that brief$/ do
+  url = "/buyers/frameworks/#{@framework_slug}/requirements/#{@lot_slug}/#{@brief_id}/award-contract"
+  page.find(:xpath, "//a[@href='#{url}']").click
+end
+
 Given /^I go to that brief overview page$/ do
   url = "/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/#{@lot_slug}/#{@brief_id}"
```
katstevens: Do you think we should use the `@framework_slug` here as well?

benvand: Yeah good shout, we can add that in later, tech debt ticket in _our_ sprint?

PR #343, opened 2017-09-04
Use a valid Companies House number

TheDoubleK:


TheDoubleK:
Smart thinking Keenoy :+1:

PR #344, opened 2017-09-04
Use a valid Companies House number for update too

TheDoubleK:


TheDoubleK:
Second time lucky :+1:

PR #345, opened 2017-09-05
Add ability to run tests in parallel

Wynndow:
Using the gem `parallel_cucumber` allows functional tests to be run in
parallel. The default setting is to run the same number of processes as
cpu cores.

The feature files are divided amongst the processes.

On my local machine the functional test suite ran in 4m30s rather than
8m30s.

This was inspired by reading this blog post.
https://tech.just-eat.com/2015/10/12/run-your-tests-faster-with-parallel-cucumber/


MuriloDalRi:
Nice!

PR #348, opened 2017-09-05
Add location to local nginx conf for direct-award

Wynndow:
The new direct award stuff sits behind `/buyers/direct-award` but lives
in the buyer app. Without this new location, requests to that URL are
sent to the briefs app.

### DIFF
--- a/nginx/docker_nginx.conf
+++ b/nginx/docker_nginx.conf
@@ -30,6 +30,12 @@ http {
             proxy_pass http://user-frontend:80;
         }
 
+        location /buyers/direct-award {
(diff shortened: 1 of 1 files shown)


katstevens:
Hello! Can we add this to the non-docker local nginx as well? I just put up a fix for the `/user` location proxy https://github.com/alphagov/digitalmarketplace-functional-tests/pull/346

Wynndow:
@katstevens I think it already has it! I'll take a look at your user app one now though.

katstevens:
Ooops yes it does, sorry (I was looking at the wrong file). Thanks!

Wynndow:
No worries! Thanks!

PR #352, opened 2017-09-07
Update content for save search button

samuelhwilliams:
## Summary
A recent update to the 'Save search' button on the G-Cloud Direct Award flow has changed the content, so we need to update it here.


samuelhwilliams:
Got Louise to 👍 this on Slack as she doesn't have access to collaborate on this yet.

PR #357, opened 2017-09-11
View your account homepage links differ on environments

idavidmcdonald:


benvand:
Need to check that the test runner will collect both of these even though they have the same name?
```
class Foo(TestCase):
    def test_foo():
        pass
    def test_foo():
        pass
```
For example the above will only collect one test in python

benvand:
This SO answer implies that the different tags will take care of it:
https://stackoverflow.com/a/37047734

benvand:
LGTM 👍

PR #358, opened 2017-09-12
Add tests for cancel and unsuccessful brief flows

CrystalPea:
This is related to following trello tickets:
Flow: https://trello.com/c/Umt24W6d/816-3-collect-status-when-buyer-has-cancelled-procurement-no-journey-frontend

Banners: https://trello.com/c/Umt24W6d/816-3-collect-status-when-buyer-has-cancelled-procurement-no-journey-frontend

### DIFF
--- a/features/buyer/cancel_requirements.feature
+++ b/features/buyer/cancel_requirements.feature
@@ -22,7 +22,7 @@ Scenario: Cancel a requirement
   When I choose the 'The requirement has been cancelled' radio button
   And I click the 'Update requirement' button
   Then I see 'The contract was not awarded' text on the page
-  Then I see 'the requirement was cancelled.' text on the page
+  And I see 'the requirement was cancelled.' text on the page
   And I see a success banner message containing 'updated'
   And I see the 'View suppliers who applied' link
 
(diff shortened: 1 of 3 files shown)


features/buyer/cancel_requirements.feature
```diff
@@ -0,0 +1,60 @@
+@cancel
+Feature: Cancel a requirement
+  In order to ensure the procurement process is fair and transparent
+  As a buyer within government
+  I want to be able to publish the result of my procurement process
```
galund: I don't see how this feature description relates to the heading!

CrystalPea: Cancelling the requirement would be a result of my procurement process, just as awarding is one, right?

galund: Sort-of, it means there isn't a procurement process. But shouldn't the "Feature" at the top of a file reflect exactly what's in the file (no more no less), in which case it should mention cancelling specifically as the user story in question? Maybe I misunderstood how these are supposed to look..

features/step_definitions/common_steps.rb
```diff
@@ -291,10 +291,10 @@
 end
 
 Then /^I (don't |)see '(.*)' text on the page/ do |negative, expected_text|
```
benvand: http://gph.is/2aNvaEd

features/step_definitions/brief_response_steps.rb
```diff
@@ -21,6 +21,9 @@
 Given /^I am logged in as the buyer of a closed brief$/ do
   closed_brief = get_briefs('digital-outcomes-and-specialists-2', 'closed').sample
   @buyer = closed_brief['users'][0]
```
benvand: Yeah this should be cool. Could have unexpected consequences if we use both
`I am logged in as the buyer of a closed brief`
`I am logged in as the buyer of a closed brief with responses`
in the same test anywhere at the moment but seems sensible.

features/buyer/cancel_requirements.feature
```diff
@@ -0,0 +1,60 @@
...
+  Then I see a temporary-message banner message containing 'This opportunity was cancelled'
+  And I see a temporary-message banner message containing 'The buyer cancelled this opportunity,'
+  And I see a temporary-message banner message containing 'for example because they no longer have the budget.'
+  And I see a temporary-message banner message containing 'They may publish an updated version later.'
+
+
+@skip-staging
+Scenario: Cancel a requirement where no suitable suppliers applied
+  Given I am logged in as the buyer of a closed brief
+
```
benvand: Might be worth having @idavidmcdonald run his eyes over this bit. I know there was a broblem aound the whole `View your account` > `View your requirements` step on preview on Monday

idavidmcdonald: The tags look good to me
When the cancelling flow gets deployed to STAGING, we will need to add new functional tests which have different links to view the list of DOS opportunities. I can explain more when you get to that point in person if you'd like (or Sam can if I'm busy)

features/buyer/cancel_requirements.feature
CrystalPea: ok, I found a more appropriate one I think: "As a buyer, I need to cancel my requirement after applications have closed, because I didn't end up going ahead with the procurement."

Is it ok? :)

features/step_definitions/brief_response_steps.rb
CrystalPea: Ahhh, interesting :o. We could actually refactor them into a single step with "with responses" optional... hmmm :d

features/buyer/cancel_requirements.feature
CrystalPea: 👍

katstevens: I've duplicated the tests, removing the new intermediate buyer account page under the `@skip-local @skip-preview` tags.

features/buyer/cancel_requirements.feature
katstevens: Now updated

galund: 👍

PR #360, opened 2017-09-14
Buyer account page

DilwoarH:
As a G-Cloud buyer
I need to be able to return to a list of the projects i am running
So that I can keep track of my projects

Ticket: https://trello.com/c/FiR3NLyZ/679-buyer-account-page

### DIFF
--- a/features/buyer/buyer_dashboard.feature
+++ b/features/buyer/buyer_dashboard.feature
@@ -10,7 +10,7 @@ Scenario: Users should see new Dashboard
 
 Scenario: Users should see link when there are searches available to view
   Given I am logged in as a buyer user
-  And I have a created and saved a search
+  And I have created and saved a search
   And I am on the /buyers page
   Then I see the 'View your saved searches' link
 
--- a/features/buyer/buyer_dashboard.feature
+++ b/features/buyer/buyer_dashboard.feature
@@ -1,0 +1,25 @@
+@skip-staging @skip-production
+@buyer-dashboard
+Feature: Buyer Dashboard
+
+Scenario: Users should see new Dashboard
+  Given I am logged in as a buyer user
+  And I am on the /buyers page
+  Then I see 'Cloud hosting, software and support' text on the page
+  And I see 'Digital outcomes, specialists and user research' text on the page
+
+Scenario: Users should see link when there are searches available to view
+  Given I am logged in as a buyer user
+  And I have a created and saved a search
+  And I am on the /buyers page
+  Then I see the 'View your saved searches' link
+
+Scenario: Users should see message when no searches are created
+  Given I am logged in as a buyer user
+  And I am on the /buyers page
+  Then I see 'You don't have any saved searches' text on the page
+
+Scenario: Users should see message when no requirements are created
+  Given I am logged in as a buyer user
+  And I am on the /buyers page
+  Then I see 'You don't have any requirements' text on the page
--- a/features/buyer/direct-award.feature
+++ b/features/buyer/direct-award.feature
@@ -4,9 +4,5 @@ Feature: Direct Award flows
 
 Scenario: User can save a search into a new Direct Award Project
   Given I am logged in as a buyer user
-  And I am on the /g-cloud/search page
-  And I click 'Save search'
-  Then I am on the 'Save your search' page
-  And I enter 'my cloud project' in the 'Name your search' field
-  And I click 'Save and continue'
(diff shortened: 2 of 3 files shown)


features/step_definitions/direct_award_steps.rb
```diff
@@ -0,0 +1,9 @@
+Given(/^I have a saved search$/) do
+  steps %Q{
+    Given I am on the /g-cloud/search page
+    And I click 'Save search'
+    Then I am on the 'Save your search' page
+    And I enter 'my cloud project' in the 'Name your search' field
+    And I click 'Save and continue'
+  }
+end
```
samuelhwilliams: Newline

features/buyer/buyer_dashboard.feature
```diff
@@ -0,0 +1,31 @@
...
+Scenario: Users should see link when there are requirements available to view
+  Given I am logged in as a buyer user
+  And I have created an individual specialist requirement
+  And I am on the /buyers page
+  Then I see the 'View your requirements' link
+
+Scenario: Users should see message when no requirements are created
+  Given I am logged in as a buyer user
+  And I am on the /buyers page
+  Then I see 'You don't have any requirements' text on the page
```
samuelhwilliams: Newline.

idavidmcdonald:
I'd agree on the unit test front. If they aren't already, we should already cover this in unit tests.

I think functional tests would be more useful if your are navigating between several pages, for example filling in a form. The tests you've written that just check the account pages might not be needed so much. The saving a search is still a good candidate though I think.

features/step_definitions/direct_award_steps.rb
DilwoarH: done

features/buyer/buyer_dashboard.feature
DilwoarH: done

features/buyer/direct-award.feature
```diff
@@ -4,9 +4,5 @@ Feature: Direct Award flows
 
 Scenario: User can save a search into a new Direct Award Project
   Given I am logged in as a buyer user
-  And I am on the /g-cloud/search page
-  And I click 'Save search'
-  Then I am on the 'Save your search' page
-  And I enter 'my cloud project' in the 'Name your search' field
-  And I click 'Save and continue'
+  And I have a saved search
   Then I am on the 'my cloud project' page
```
samuelhwilliams: This test doesn't really make sense with the steps that it ends up with. It should have the steps in there properly so that it reads like a complete scenario. It might create some duplication, but it just doesn't make sense as it is: `Given I am logged in as a buyer, And I have a saved search, Then I am on the 'my cloud project' page.`

galund: Agree, I think if you really want to deduplicate then you might come up with a better name for the shared step, perhaps instead of `Given(/^I have a saved search$/) ` you could have `When(/^I create and save a search$/) `

features/buyer/buyer_dashboard.feature
```diff
@@ -0,0 +1,31 @@
...
+  And I have a saved search
+  And I am on the /buyers page
+  Then I see the 'View your saved searches' link
+
+Scenario: Users should see message when no searches are created
+  Given I am logged in as a buyer user
+  And I am on the /buyers page
+  Then I see 'You don't have any saved searches' text on the page
+
+Scenario: Users should see link when there are requirements available to view
```
galund: Yeah, so per @samuelhwilliams's earlier comment, I think this test is entirely covered by the existing test in `requirements.feature` called "View requirement in a dashboard". This doesn't test anything that doesn't test already, as far as I can see.

So let's remove the duplicative ones like this, and see what we're left with. Testing the case where there's no requirements / no G-Cloud projects is okay, I guess.

And we need to make sure we've got a G-Cloud equivalent of that DOS test, i.e. one where you actually click through all the way from the home page to right down to the task list.

And somewhere we want a test that takes a project right through all the stages, i.e. End Search and Download. That could be here too, unless it's already somewhere else and I've missed it? cheers

DilwoarH:
@samuelhwilliams @galund comments have been addressed

features/step_definitions/direct_award_steps.rb
```diff
@@ -0,0 +1,9 @@
+When(/^I have a created and saved a search$/) do
```
galund: Extra "a"

PR #362, opened 2017-09-14
add subscribe_mailing_list.feature

risicle:
This starts its journey at `/suppliers/mailing-list` because the links leading to it aren't in place yet, but the intention is for it to follow the links from the homepage once they are.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -192,6 +192,10 @@ Then /^I see a (success|warning|destructive|temporary-message) banner message co
   banner_message.should have_content(message)
 end
 
+Then /^I don't see a banner message$/ do
+  page.should_not have_selector(:xpath, "//*[contains(@class, 'banner-')][contains(@class, '-action')]")
+end
+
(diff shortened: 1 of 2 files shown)


risicle:
Yup this was discussed, see the trello card. It's not possible to distinguish between "this looks like a phoney address" and "you've blacklisted your email from us" as they are signalled in the same way by MC. It was decided against trying to disentangle all the possible failure modes and write content for each obscure case.

risicle:
Tags? I have no idea. I thought only the `@smoke-test`s got run on production?

TheDoubleK:
OK probably the tags are a "legacy" thing then I think, ignore me.

:+1:

risicle:
No really I don't know, we should get a proper answer on this before merging.

risicle:
Tags are ok apparently.

TheDoubleK:
Actually don't merge this now after all.  The message it checks for isn't quite right (it has the old full stop in).  I've pulled this commit and fixed it in my branch locally so will make a new PR with full flow tests.

PR #363, opened 2017-09-14
Reset password email tests

allait:
This adds step definitions for flows that include sending an email to the user.

By making notification references predictable we can use Notify API to retrieve the email by reconstructing the reference that the application attached to it in the functional test code.

This allows us to get the email body directly from Notify.

The benefit of this approach is that we can use test Notify API keys and avoid sending actual emails to fake email addresses. This is also a much simpler setup since it doesn't require setting up an email address to receive emails or adding delays to the tests to account for email delivery.


features/step_definitions/emails.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+end
+
+Then /^I receive a '([a-z-]+)' email for #{MAYBE_VAR}/ do |message_type, email_address|
+  messages = DMNotify.get_email(message_type, email_address)
+  @email_text = messages.collection[0].body
+end
+
+Then /^I click the link in that email$/ do
+  page.visit(URI.extract(@email_text).select { |i| i.start_with?('http') } [0])
```
samuelhwilliams: Is it not going to always start with 'https'?

features/user/reset_password.feature
```diff
@@ -0,0 +1,17 @@
+@reset-password @notify
+Feature: Reset user password
+
+Scenario: Request password reset
+  Given I have a buyer user
+  And I am on the /user/reset-password page
+  When I enter that user.emailAddress in the 'Email address' field
+  And I click 'Send reset email' button
+  Then I see a success banner message containing 'send a link to reset the password'
```
samuelhwilliams: Hmm, it feels like the content might have been better as `If the email address you've entered belongs to a Digital Marketplace account, we've sent a link to reset the password.` ... But I guess that's a different matter.

features/step_definitions/emails.rb
allait: Not when run locally

samuelhwilliams: Ah ok 👍

PR #371, opened 2017-09-19
Add a `wait` step to reset email password scenario

Wynndow:
Preview notify tests are too quick. We create a user and request
password reset immediately. It looks like the date fernet puts inside
the token is stored with a second resolution (so it doesn’t have
milli/microseconds), but the API response does. So if the user creation
and token generation happen within the same second (which it looks like
they do quite often) the check thinks that the token was created before
the user and declares it expired.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -312,3 +312,7 @@ And /^I wait for the page to reload/ do
     loop until page.evaluate_script('jQuery.active').zero?
   end
 end
+
+And /^I wait (\d+) seconds/ do |seconds|
+  sleep seconds.to_i
+end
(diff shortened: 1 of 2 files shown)


features/user/reset_password.feature
```diff
@@ -8,6 +8,7 @@ Scenario: Request password reset
   And I click 'Send reset email' button
   Then I see a success banner message containing 'send a link to reset the password'
   And I receive a 'reset-password' email for that user.emailAddress
+  And I wait 2 seconds
```
samuelhwilliams: It might be nice to add an optional inline comment to this step so that you can give some explanation about why, i.e. `And I wait 2 seconds to ensure the reset token is valid`

features/step_definitions/common_steps.rb
```diff
@@ -312,3 +312,7 @@
     loop until page.evaluate_script('jQuery.active').zero?
   end
 end
+
+And /^I wait (\d+) seconds/ do |seconds|
```
samuelhwilliams: It might be nice to add an optional inline comment to this step so that you can give some explanation about why, i.e. `And I wait 2 seconds to ensure the reset token is valid`

Wynndow: Nice. Have added.

PR #375, opened 2017-09-20
Intentionally break smoke tests so we can test jenkins job

idavidmcdonald:
Our jenkins job is not alerting when smoke tests fail. By introducing
this failure, we can then change our jenkins job until the errors
start appearing again on slack. Once the errors are back then this
commit can be reverted.


MuriloDalRi:
👍

risicle:
*head in hands*

PR #376, opened 2017-09-21
Add staging/production variant of smoketest

TheDoubleK:
Text is changing on the homepage link, but we need the smoketest to continue running on production while this new feature is tested on staging.

So this should be a very temporary stop-gap to keep the board green.

### DIFF
--- a/features/smoke-tests/buyer/catalogue.feature
+++ b/features/smoke-tests/buyer/catalogue.feature
@@ -2,6 +2,7 @@
 @catalogue
 Feature: Passive catalogue buyer journey
 
+@skip-preview @skip-staging @skip-local
 Scenario: User can see the main links on the homepage
   Given I am on the homepage
   Then I see the 'Find cloud hosting, software and support' link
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
One of these will also need a `@skip-development` tag, depending on which one is most likely to be around on dev config. Otherwise they'll both run locally.

TheDoubleK:
Who runs things locally? :trollface:

samuelhwilliams:
One of these will also need a `@skip-local` tag, depending on which one is most likely to be around on dev config. Otherwise they'll both run locally.

TheDoubleK:
This morning?  Because it'll be gone after that.

TheDoubleK:
And who knows what branches devs have checked out locally?

TheDoubleK:
Done.

TheDoubleK:
If someone approves this please go ahead and merge ASAP to fix preview - I have to go to school now.

PR #379, opened 2017-09-22
Move locked account functional test scenario from legacy to master

TheDoubleK:
There are more to go before we can completely delete the "Legacy" tests here: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/supplier/supplier_journey.feature

But... it's the end of the mission so I'm going to PR these on their own now.  Slowly slowly we will whittle it away...

I also adjusted the password reset Scenario to perform the journey from the homepage, rather than just starting directly at the page itself.

This feels like a useful thing for functional tests to do - we want reassurance that links between different frontend apps are all working OK, and that users can do the thing without needing to know a particular URL.

### DIFF
--- a/features/user/reset_password.feature
+++ b/features/user/reset_password.feature
@@ -1,10 +1,18 @@
 @reset-password @notify @skip-staging
 Feature: Reset user password
 
-Scenario: Request password reset
+Background:
   Given I have a buyer user
   And I wait 2 seconds to ensure the reset token is created after the user
-  And I am on the /user/reset-password page
+
+Scenario: User has forgotten their password and requests a password reset
+  When I am on the homepage
+  And I click 'Log in'
+  Then I am on the 'Log in to the Digital Marketplace' page
+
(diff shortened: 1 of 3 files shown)


features/user/reset_password.feature
```diff
@@ -1,10 +1,18 @@
...
 Feature: Reset user password
 
-Scenario: Request password reset
+Background:
   Given I have a buyer user
   And I wait 2 seconds to ensure the reset token is created after the user
-  And I am on the /user/reset-password page
+
+Scenario: User has forgotten their password and requests a password reset
+  When I am on the homepage
```
risicle: Shouldn't this also be a `Given`? (does it make any difference?)

risicle:
WFM :+1:

features/user/reset_password.feature
TheDoubleK: Probably (no it doesn't make any difference)

PR #389, opened 2017-10-04
Rewrite test for buyer creating an account 

CrystalPea:
Legacy test: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/buyer/buyer_creation.feature

Trello ticket: https://trello.com/c/6EXOJr2h/43-1-functional-test-create-buyer-account


features/step_definitions/emails.rb
```diff
@@ -23,3 +23,9 @@ def self.get_email(message_type, email_address)
 Then /^I click the link in that email$/ do
   page.visit(URI.extract(@email_text).select { |i| i.start_with?('http') } [0])
 end
+
+Given /^I have an email address$/ do
```
TheDoubleK: Maybe this should be "I have an email address with an allowed buyer email domain" to make it clearer?

features/buyer/buyer_creation.feature
```diff
@@ -0,0 +1,27 @@
+@newbuyer @skip-staging @notify
+Feature: Create buyer account. We're assuming a user journey where someone tries to find a specialist, eventually hits the login page, and then clicks 'Create a buyer account' from the login page
```
TheDoubleK: I don't think we need the second sentence here - the flow is clear from the steps taken. It feels like the "Feature" and "Scenario" are the same thing here, but if we want to be specific then could go with something like:
```
Feature: Create a buyer account

Scenario: Create a new buyer account when trying to create a requirement from the home page
```

features/step_definitions/emails.rb
CrystalPea: Good idea to signify that it's a buyer-allowed email address, I will change it now.

features/buyer/buyer_creation.feature
CrystalPea: That's a very good point to put additional details in scenario rather than feature, then we could potentially have different journey for buyer to create account in the same feature in the future :).

PR #404, opened 2017-11-01
Clean up functional tests

DilwoarH:
Check that functional tests cover all possible flows

Scenarios:
----------------
- Unauthenticated user can save a search after logging in
- User can save a search
- User with saves searches completes new saved search
- User updates existing saved search
- User edits existing search
- User ends search and downloads results
- User redownloads results

Journeys:
----------------
- User not logged in -> goes to g-cloud search -> gets presented login screen -> logs in -> saves the search -> sees project page.
- User logged in -> goes to g-cloud search -> saves the search -> sees project page.
- User logged in > has saved searches -> goes to g-cloud search -> saves search -> creates new saved search -> sees project page.
- User logged in > has saved searches -> goes to g-cloud search -> saves search -> overrides existing saved search -> sees project page.
- User logged in > goes to my accounts > view saved searches page > chooses a search > edits a search > updates a search
- User logged in > goes to my accounts > view saved searches page > chooses a search > ends search > downloads results
- User logged in > goes to my accounts > view saved searches page > chooses a closed search > downloads results

Ticket: https://trello.com/c/ctU5jhzr/56-clean-up-functional-tests

### DIFF
--- a/features/buyer/direct-award.feature
+++ b/features/buyer/direct-award.feature
@@ -1,7 +1,85 @@
 @direct-award
 Feature: Direct Award flows
-
-Scenario: User can save a search into a new Direct Award Project
-  Given I am logged in as a buyer user
-  And I have created and saved a search
+  
+Scenario: Unauthenticated user can save a search after logging in
+  Given I have a production buyer user
+  And I am on the /g-cloud/search page
+  And I click 'Save search'
+  Then I am on the 'Log in to the Digital Marketplace' page
+  When I enter that user.emailAddress in the 'Email address' field
+  And I enter that user.password in the 'Password' field
+  And I click the 'Log in' button
+  Then I see the 'Log out' link
+  Then I am on the 'Choose where to save your search' page
+  And I choose the 'Create a new saved search' radio button
+  And I enter 'my cloud project' in the 'Name your search' field
+  And I click 'Save and continue'
   Then I am on the 'my cloud project' page
+
+Scenario: User can save a search
+  Given I am logged in as a buyer user
+  And I am on the /g-cloud/search page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I enter 'my cloud project' in the 'Name your search' field
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project' page
+
+Scenario: User with saved searches completes new saved search
+  Given I am logged in as a buyer user
+  And I have created and saved a search called 'my cloud project'
+  And I am on the /g-cloud/search page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I choose the 'Create a new saved search' radio button
+  And I enter 'my cloud project2' in the 'Name your search' field
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project2' page
+
+Scenario: User updates existing saved search
+  Given I am logged in as a buyer user
+  And I have created and saved a search called 'my cloud project - exiting'
+  And I am on the /g-cloud/search page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I choose the 'my cloud project - exiting' radio button
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project - exiting' page
+
+Scenario: User edits existing search
+  Given I am logged in as a buyer user
+  And I have created and saved a search called 'my cloud project'
+  And I am on the /buyers page
+  Then I click the 'View your saved searches' link
+  Then I click the 'my cloud project' link
+  Then I am on the 'my cloud project' page
+  And I click the 'Edit search' link
+  Then I am on the 'Search results' page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I choose the 'my cloud project' radio button
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project' page
+
+Scenario: User ends search and downloads results
+  Given I am logged in as a buyer user
+  And I have created and saved a search called 'my cloud project'
+  And I am on the /buyers page
+  Then I click the 'View your saved searches' link
+  Then I click the 'my cloud project' link
+  Then I am on the 'my cloud project' page
+  And I click the 'End search' link
+  Then I am on the 'End your search' page
+  And I click the 'End search and continue' button
+  Then I am on the 'my cloud project' page
+  And I click the 'Download search results' link
+  And I am on the 'Download your search results' page
+  And I click the 'Download search results as a spreadsheet' link
+  And I should get a download file of type 'ods'
+  And I click the 'Return to overview' link
+  Then I am on the 'my cloud project' page
+  And I click the 'Download your results again.' link
+  And I am on the 'Download your search results' page
+  And I click the 'Download search results as a spreadsheet' link
+  And I should get a download file of type 'ods'
+ 
--- a/features/step_definitions/direct_award_steps.rb
+++ b/features/step_definitions/direct_award_steps.rb
@@ -1,9 +1,9 @@
-When(/^I have created and saved a search$/) do
+When(/^I have created and saved a search called '(.*)'$/) do |search_name|
   steps %Q{
-    Given I am on the /g-cloud/search page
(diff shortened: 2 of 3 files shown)


features/step_definitions/direct_award_steps.rb
```diff
@@ -1,9 +1,14 @@
...
+    Given I am on the /g-cloud/search?q=test+123 page
     And I click 'Save search'
     Then I am on the 'Choose where to save your search' page
     And I enter 'my cloud project' in the 'Name your search' field
     And I click 'Save and continue'
   }
 end
+
+
+Then(/^I should get a ods download file$/) do
```
galund: "an ODS" pls :)

features/step_definitions/direct_award_steps.rb
```diff
@@ -1,9 +1,14 @@
...
+  expect(page.response_headers['Content-Disposition']).to match(/attachment;filename=[\s\S]*.ods/)
```
galund: So we maybe reckon `filename=\S*\.ods` would do the trick

DilwoarH:
@galund updated

features/buyer/direct-award.feature
```diff
@@ -1,7 +1,103 @@
 @direct-award
 Feature: Direct Award flows
+  
+Scenario: Unauthenticated user can save a search after logging in
+  Given I have a production buyer user
+  And I am on the /g-cloud/search page
```
TheDoubleK: All of these scenarios have the user saving an empty search to start with.  Might be nice to have some where they enter keywords, click some filters and then save a new search.

features/step_definitions/direct_award_steps.rb
```diff
@@ -1,9 +1,14 @@
 When(/^I have created and saved a search$/) do
   steps %Q{
-    Given I am on the /g-cloud/search page
+    Given I am on the /g-cloud/search?q=test+123 page
```
TheDoubleK: "test 123" is a search with zero results (at least with data as it currently stands) - feels like there's a danger here that we never test downloading an ODS file with any rows in it - might be better to pick base search terms that will return some results.

There is then the problem of ensuring that the final "download results" isn't too big... `/g-cloud/search?q=email+analysis+provider` has 19 results on prod at the moment, could work?

Also, do we want an extra test for where a user can't download too many results (e.g. search for `software` with no filters and check the expected error message appears)?

features/step_definitions/direct_award_steps.rb
```diff
@@ -1,9 +1,14 @@
...
     And I click 'Save search'
     Then I am on the 'Choose where to save your search' page
     And I enter 'my cloud project' in the 'Name your search' field
     And I click 'Save and continue'
   }
 end
+
+
+Then(/^I should get an ODS download file$/) do
```
TheDoubleK: This could be a parameterised general-purpose step in `common_steps.rb` - would be useful for testing CSV downloads, etc.

TheDoubleK: This is how it is done in the "Legacy" branch: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/features/step_definitions/journey_test_steps.rb#L1329

I'm not saying that's the best way, but might be worth a look.

TheDoubleK: But if we don't know the exact filename or expected file size it doesn't look like there's much extra there to be honest.

features/step_definitions/direct_award_steps.rb
DilwoarH: Thanks @TheDoubleK - Will update the query.

Regarding the "too many results" flow. We have tested this in the unit tests. So I think that will be fine.

features/step_definitions/direct_award_steps.rb
DilwoarH: I'll move this to common_steps.rb but we don't know the filename or expected size so therefore I think this is probably still the best way of doing it.

DilwoarH:
@TheDoubleK - your comments have been addressed.

features/buyer/direct-award.feature
```diff
@@ -1,7 +1,103 @@
...
+  And I click the 'End search' link
+  Then I am on the 'End your search' page
+  And I click the 'End search and continue' button
+  Then I am on the 'my cloud project' page
+  And I click the 'Download search results' link
+  And I am on the 'Download your search results' page
+  And I click the 'Download search results as a spreadsheet' link
+  And I should get a download file of type 'ods'
+  And I click the 'Return to overview' link
+  And I take a screenshot
```
TheDoubleK: We don't need the "take a screenshot" step in here.

features/buyer/direct-award.feature
```diff
@@ -1,7 +1,103 @@
...
+  And I click the 'End search' link
+  Then I am on the 'End your search' page
+  And I click the 'End search and continue' button
+  Then I am on the 'my cloud project' page
+  Then I click the 'Download search results' link
+  And I am on the 'Download your search results' page
+  And I click the 'Download search results as a spreadsheet' link
+  And I should get a download file of type 'ods'
+
+@direct-award2
```
TheDoubleK: I'm not sure we need a tag just for this scenario?  The tag for the feature should be enough.

features/buyer/direct-award.feature
```diff
@@ -1,7 +1,103 @@
...
+  Then I am on the 'my cloud project' page
+  And I click the 'Edit search' link
+  Then I am on the 'Search results' page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I choose the 'my cloud project' radio button
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project' page
+
+Scenario: User ends search and downloads results
```
TheDoubleK: As all the steps here are duplicated in the following Scenario I think we can drop this first one altogether, and rename the following one as something like "User can end their search, download results and return to re-download them later."

features/buyer/direct-award.feature
```diff
@@ -1,7 +1,103 @@
...
+Scenario: User can save a search
+  Given I am logged in as a buyer user
+  And I am on the /g-cloud/search page
+  And I click 'Save search'
+  Then I am on the 'Choose where to save your search' page
+  And I enter 'my cloud project' in the 'Name your search' field
+  And I click 'Save and continue'
+  Then I am on the 'my cloud project' page
+
+Scenario: User with saves searches completes new saved search
```
TheDoubleK: `saves` -> `saved`

features/step_definitions/direct_award_steps.rb
```diff
@@ -1,6 +1,6 @@
 When(/^I have created and saved a search$/) do
```
TheDoubleK: It might be nice to parameterise this with the name to save as, which means the Scenarios can set up searches with different names, e.g. `I have created and saved a search called 'my other cloud project'`

It could even use the name passed in as the search query too, e.g. `I have created and saved a search for 'email analysis provider'`' - and then different scenarios can use different queries and have searches saved with different names.  This multi-step method would then be less of a black box in terms of what it creates for the Scenarios (at the moment the Scenarios assume 'my cloud project' as a name, but it isn't obvious from reading the scenarios on their own where that name comes from - by parameterising it I think it will make the scenario files easier to follow.

features/buyer/direct-award.feature
DilwoarH: the first one the user doesn't have any saved searches, the second one the user has existing searches

DilwoarH:
@TheDoubleK - updated all of them but one (left comments).

features/buyer/direct-award.feature
TheDoubleK: I can't see a difference - all the scenario steps are exactly the same, so all the data setup must be the same? i.e. the 14 steps of `Scenario: User ends search and downloads results` are the **exact same 14 steps** at the start of `Scenario: User redownloads results` - so these 14 steps are performed twice exactly the same.

DilwoarH: Hi @TheDoubleK - I have merged the two together.

TheDoubleK:
Looks good now, but please can you tidy up the commit history (either squash "updates" into previous commit, or add a more descriptive commit message).  Will be good to go after that.

PR #406, opened 2017-11-06
Duplicate ODS download check; one of them should be CSV

samuelhwilliams:
## Summary
One of the functional tests downloads saved search results, but downloads the ODS twice rather than the CSV on the second time.


galund:
jinks @TheDoubleK

samuelhwilliams:
Interesting. Couldn't tell that was the case from the test name. Either way, we recently had a problem where one of the downloads worked but the other didn't... so let's get this in regardless.

PR #407, opened 2017-11-09
Iterate through brief responses until response is found

idavidmcdonald:
We have run into problems where we call `get_brief_responses` but
end up with no responses with the required brief status. By changing
it so we can essentially pick from more than the first 100 results,
we can make it much more unlikely that we won't be able to find
a suitable response.

Tests pass on preview which currently has closed briefs included in the first 100 brief responses. Tests also pass on my local which has no closed briefs included in the first 100 brief responses.

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -167,7 +167,19 @@ def get_brief_responses(framework_slug, brief_response_status, brief_status)
   params = {status: brief_response_status, framework: framework_slug}
   response = call_api(:get, '/brief-responses', params: params)
   brief_response_list = JSON.parse(response.body)['briefResponses']
-  brief_response_list.select { |x| x['brief']['status'] == brief_status }
+  brief_response_list = brief_response_list.select { |x| x['brief']['status'] == brief_status }
+
+  # If we have not found any required brief responses so far on the first page then we will iterate
+  # through the pages until we do find atleast one
+  next_page = 2;
+  while brief_response_list.empty? && JSON.parse(response.body)['links']['next']
+    response = call_api(:get, "/brief-responses?page=#{next_page}", params: params)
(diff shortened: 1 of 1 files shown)


features/support/api_helpers.rb
```diff
@@ -167,7 +167,19 @@ def get_brief_responses(framework_slug, brief_response_status, brief_status)
...
   response = call_api(:get, '/brief-responses', params: params)
   brief_response_list = JSON.parse(response.body)['briefResponses']
-  brief_response_list.select { |x| x['brief']['status'] == brief_status }
+  brief_response_list = brief_response_list.select { |x| x['brief']['status'] == brief_status }
+
+  # If we have not found any required brief responses so far on the first page then we will iterate
+  # through the pages until we do find atleast one
+  next_page = 2;
+  while brief_response_list.empty? && JSON.parse(response.body)['links']['next']
+    response = call_api(:get, '/brief-responses?page=' + next_page.to_s, params: params)
```
CrystalPea: Here you could do string interpolation, like this: `"/brief-responses?page=#{next_page}"`
It's important to use double quotes. If you do this, then you don't have to use `.to_s`.

But of course it will work either way ☺️

features/support/api_helpers.rb
```diff
@@ -167,7 +167,19 @@ def get_brief_responses(framework_slug, brief_response_status, brief_status)
...
+    brief_response_list = JSON.parse(response.body)['briefResponses']
+    brief_response_list = brief_response_list.select { |x| x['brief']['status'] == brief_status }
+    next_page = next_page.next
```
CrystalPea: here you can also use `next_page += 1`, but either way is fine :)

idavidmcdonald: Yeah that is nicer and looks more like python 🐍  Will change.

PR #408, opened 2017-11-09
Supplier edits brief responses

katstevens:
Trello ticket: https://trello.com/c/kQpbazwt/85-allow-suppliers-edit-submitted-application-if-opportunity-is-still-open

- Splits existing features into old and new flows (using `@skip-local` / `@skip-staging` etc)
- New scenarios contain copy changes, Check Your Answers page, different flows
- Brand new scenario for editing a submitted application

We also added a new step definition for clicking on the first of a set of matching links.

To be merged once the following PRs are on preview:
- Brief Responses FE https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/11
- Buyer FE https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/659

Follow-up PR to remove the duplicate scenarios and `@skip-staging` tags once the above are on staging: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/409

### DIFF
--- a/features/supplier/supplier_applies_for_an_opportunity.feature
+++ b/features/supplier/supplier_applies_for_an_opportunity.feature
@@ -37,6 +37,8 @@ Scenario: Supplier is not eligible as they can not provide the developer role
   And I see 'You can’t apply for this opportunity because you didn’t say you could provide this specialist role when you applied to the Digital Outcomes and Specialists 2 framework.' text on the page
   And I see a 'data-reason' attribute with the value 'supplier-not-on-role'
 
+
+@skip-local @skip-preview
 Scenario: Supplier applies for a digital-specialists brief
   Given that supplier has applied to be on that framework
   And we accept that suppliers application to the framework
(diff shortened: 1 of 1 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -132,6 +132,10 @@
   page.all("." + button_link_class)[0].click
 end
 
+When /I click a #{MAYBE_VAR} link$/ do |link_name|
```
TheDoubleK: Does the step above not work for this? It seems like it should?
ie the step defined as `When /I click #{MAYBE_VAR} ?(button|link)?$/ do |button_link_name, elem_type|`

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -284,6 +475,119 @@ Scenario: Supplier changes their answers before submission
...
+  And that supplier has a service on the digital-specialists lot
+  And I have a live digital-specialists brief
+  And that supplier has filled in their response to that brief but not submitted it
+  And I go to that brief page
+  And I click 'Apply for this opportunity'
+  Then I am on 'Before you start' page
+
+  When I click 'Continue application'
+  Then I am on 'When is the earliest the specialist can start work?' page
+  And I see 'The buyer needs the specialist to start: Saturday 31 December 2016' replayed in the question advice
```
TheDoubleK: I wonder if we should bring these dates forward a bit so the tests feel fresh and relevant to now... Probably not worth the hassle though!

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -284,6 +475,119 @@ Scenario: Supplier changes their answers before submission
...
+
+  When I click the 'Tea drinker' link
+  Then I am on the 'Check your answers' page
+  When I click a 'Edit' link
+
+  Then I am on the 'What’s the specialist’s day rate?' page
+  And I see '£200' replayed in the question advice
+  And I see '200' as the value of the 'dayRate' field
+
+  When I enter '100' in the 'dayRate' field
```
TheDoubleK: It might be nice to update a few more fields (one of each type maybe, or even maybe all of them) and then check the table for all updated values.

features/supplier/supplier_applies_for_an_opportunity.feature
```diff
@@ -329,7 +633,25 @@ Scenario: Supplier can see the empty string on the opportunities dashboard
   And I click the 'View your opportunities' link
   Then I see 'You haven’t applied to any opportunities' text on the page
 
-@opportunities-dashboard
+
+@skip-local @skip-preview @opportunities-dashboard
+Scenario: Supplier can see their opportunity response
```
TheDoubleK: I think the steps here are an exact duplicate of the steps at the start of the "changes their answers" scenario? No need to duplicate these as a separate thing, it will just add to win time and test no additional things.

features/step_definitions/common_steps.rb
katstevens: Annoyingly it gives an error if there's more than one matching element on the page. Our new step selects the first match.

features/supplier/supplier_applies_for_an_opportunity.feature
katstevens: The whole point of this scenario was to check that editing a single question in the flow redirects straight back to the Check Your Answers page - if we fill out every question then it would go there by default 😸

features/supplier/supplier_applies_for_an_opportunity.feature
katstevens: Good spot! I wonder why this was ever a separate scenario? Happy to tack the final couple of steps (view opportunities link / and I see the closing date etc ) on to the new scenarios instead.

features/supplier/supplier_applies_for_an_opportunity.feature
TheDoubleK: Ah OK, makes sense.

features/step_definitions/common_steps.rb
TheDoubleK: Ah yes, I made a step for this not too long ago - see https://github.com/alphagov/digitalmarketplace-functional-tests/blob/13b0d7b97bce822cb1943fc341c782b66f3901c5/features/step_definitions/common_steps.rb#L272

`I click the summary table Edit link for '#{field_to_edit}'` - should hopefully work for this summary table.  Maybe use that if we can rather than introduce this new step definition? 

(If it doesn't "just work" though then fine to add this).

katstevens: Oooh nice - I'll give that a go. Thanks Kev!

PR #411, opened 2017-11-20
Increase max allowed upload size

TheDoubleK:
The default max upload size is small (<5MB) so we can't properly test
upload size validation locally as nginx errors before our apps.

Increasing the size here allows for a more reasitic dev environment.


TheDoubleK:
The apps block at 5MB.  Beyond that I don't really know.

PR #415, opened 2017-11-27
Update LICENSE

TheDoubleK:
MIT License thing was OK here, but we're not GOV.UK...


TheDoubleK:
Yeah I wasn't sure about that.

PR #420, opened 2017-12-04
NEW: (Admin) Functional test for adding a buyer domain

benvand:


features/admin/add_buyer_domain.feature
```diff
@@ -0,0 +1,30 @@
+@add-buyer-domain
+Feature: Admin buyer domains
+
+Scenario: Admin user can log in and add a buyer domain
+  Given I am logged in as the production admin user
+  And I am on the /admin page
+  And I click the 'Add a buyer email domain' link
+  When I enter 'test.com' in the 'new_buyer_domain' field
```
TheDoubleK: How do we guarantee that this exists?  It would be good to not rely on pre-existing data set up in our functional tests; that was one of the main reasons for rewriting from scratch last time.

features/admin/add_buyer_domain.feature
```diff
@@ -0,0 +1,30 @@
...
+  Given I am logged in as the production <role> user
+  And I am on the /admin/buyers/add-buyer-domains page
+  Then I am on the 'You don’t have permission to perform this action' page
+
+  Examples:
+    | role                    |
+    | admin-framework-manager |
+    | admin-ccs-sourcing      |
+    | admin-manager           |
+    | admin-ccs-category      |
```
TheDoubleK: Missing line terminator

features/admin/add_buyer_domain.feature
```diff
@@ -0,0 +1,30 @@
+@add-buyer-domain
+Feature: Admin buyer domains
+
+Scenario: Admin user can log in and add a buyer domain
```
TheDoubleK: This doesn't really describe the scenario that follows - where a user **can't** add a domain because it already exists.  It would be nice to have a separate step where a new domain is successfully added... maybe <a long-ish random string>.com would work for the success scenario?

features/admin/add_buyer_domain.feature
```diff
@@ -0,0 +1,30 @@
...
+
+Scenario: Admin user can see invalid format message for invalid domain string
+  Given I am logged in as the production admin user
+  And I am on the /admin page
+  And I click the 'Add a buyer email domain' link
+  When I enter 'test' in the 'new_buyer_domain' field
+  And I click the 'Add' button
+  Then I see a destructive banner message containing 'not a valid format'
+
+Scenario Outline: Correct users cannot access the add a buyer domain
```
TheDoubleK: Same again, do we need this or are we happy that unit test coverage is enough?  I'm not sure.

features/admin/add_buyer_domain.feature
benvand: Changed the wording here

features/admin/add_buyer_domain.feature
benvand: We can guarantee gov.uk exists as it is propagated from production :man_shrugging: Changed it to that.

features/admin/add_buyer_domain.feature
benvand: Done

benvand:
Yeah it's a tricky one. I'm also trying desperately to make do with the steps we already have and avoid creating loads of 'I enter a random email in the new domain textbox' sort of steps or updating our existing ones to do weird stuff. Not sure I'm going to be able to avoid it when I get on to the service ones :(

PR #422, opened 2017-12-05
NEW: (Admin) 3 x Functional tests for services

benvand:


features/admin/view_services.feature
```diff
@@ -0,0 +1,26 @@
+@view-a-service
+Feature: View a service
+
+Scenario Outline: Correct user can access the service page
+  Given I am logged in as the production <role> user
+  # TODO GET ME HERE FROM THE LOGIN
+  And I am on the /admin/services/1123456789012346 page
```
TheDoubleK: How do we know this exists? Would be nice to not rely on pre-existing data, because in theory the data could change at any point.  The tests should set up everything they need to start with - this is why we made the old tests - which relied on pre-existing data - "legacy".

features/admin/view_services.feature
```diff
@@ -0,0 +1,26 @@
...
+
+  Examples:
+    | role               |
+    | admin              |
+    | admin-ccs-category |
+
+Scenario Outline: Correct users cannot access the service page
+  Given I am logged in as the production <role> user
+  # TODO GET ME HERE FROM THE LOGIN
+  And I am on the /admin/services/1123456789012346 page
```
TheDoubleK: In this case I suppose we don't need a real service ID as users should just get blocked before it even hits the API for a service.  So the one in the scenario here could be obviously fake maybe?

features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
+@edit-a-service
+Feature: Edit a service
+
+Scenario: Only Service Manager can edit services
+  Given I am logged in as the production admin-ccs-category user
+  # TODO get me here from the search
+  #  And I have a random g-cloud service from the api
+  #  And I am on the /admin/services/@service['id'] page
+  And I am on the /admin/services/1123456789012346 page
+  And I see '£100 to £1234 per person per week' text on the page
```
CrystalPea: Is this amount always going to be the same? Is it part of our setup maybe?

features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
...
+  When I click the top-level summary table Edit link for the section 'Pricing'
+  And I enter '1000' in the 'Minimum price' field
+  And I click the 'Save and return to summary' button
+  Then I see '£1000 to £1234 per person per week' text on the page
+  When I click the top-level summary table Edit link for the section 'Pricing'
+  And I enter '100' in the 'Minimum price' field
+  And I click the 'Save and return to summary' button
+  Then I see '£100 to £1234 per person per week' text on the page
+
+Scenario Outline: Correct users cannot edit services
```
CrystalPea: Maybe instead of "Correct users" we could put "Admins other than Service Manager"? It would tell us more about the context I think.

features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
...
+  Given I am logged in as the production <role> user
+  When I am on the /admin/services/1123456789012346 page
+  Then I don't see the 'Edit' link
+  When I am on the /admin/services/1123456789012346/edit/pricing page
+  And I take a screenshot
```
CrystalPea: This can be taken out I guess :)

PR #424, opened 2017-12-05
NEW: (Admin) Functional tests for removing a service

benvand:


features/admin/remove_services.feature
```diff
@@ -0,0 +1,28 @@
+@remove-a-service
+Feature: Remove a service
+
+Scenario: Only Service Manager can remove services
+  Given I am logged in as the production admin-ccs-category user
+  # TODO GET ME HERE FROM THE LOGIN
+  And I am on the /admin/services/1123456789012346 page
```
TheDoubleK: This is another one where we shouldn't rely on a particular service already existing - should set up all the data we need for the test in a "Backgound" step if possible.

features/admin/remove_services.feature
```diff
@@ -0,0 +1,28 @@
...
+  Then I don't see the 'Remove service' link
+
+  Examples:
+    | role                    |
+    | admin-framework-manager |
+    | admin-ccs-sourcing      |
+    | admin-manager           |
+    | admin                   |
+
```
TheDoubleK: Extra empty line... probably fine 😄

features/admin/remove_services.feature
```diff
@@ -0,0 +1,29 @@
+@remove-a-service
+Feature: Remove a service
+
+Scenario: Only Service Manager can remove services
+  Given I am logged in as the production admin-ccs-category user
+  # TODO get me here from the search
```
TheDoubleK: Maybe for this we just make a new admin step (i.e. in `admin_steps.rb`) - something like `And I am on a random g-cloud service page` - that calls `I have a random g-cloud service from the api` internally and navigates to the page?

PR #425, opened 2017-12-06
NEW: (Admin) Functional tests for edit services

benvand:


features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
...
+  And I click the 'Save and return to summary' button
+  Then I see '£100 to £1234 per person per week' text on the page
+
+Scenario Outline: Correct users cannot edit services
+  Given I am logged in as the production <role> user
+  When I am on the /admin/services/1123456789012346 page
+  Then I don't see the 'Edit' link
+  When I am on the /admin/services/1123456789012346/edit/pricing page
+  And I take a screenshot
+  Then I am on the 'You don’t have permission to perform this action' page
```
benvand: dont need the screenshot

features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
+@edit-a-service
+Feature: Edit a service
+
+Scenario: Only Service Manager can edit services
+  Given I am logged in as the production admin-ccs-category user
+  # TODO get me here from the search
```
TheDoubleK: Same as with #424, I reckon a new admin step `I am on a random g-cloud service page` will do the job here.  Although then we can't so easily check the initial values without patching up something ugly like `I see '£that service.minPrice to £that service.maxPrice per that service.unit per that service.interval' text on the page` (I am NOT suggesting we do this!!!)

features/admin/edit_a_service.feature
```diff
@@ -0,0 +1,33 @@
...
+  And I click the 'Save and return to summary' button
+  Then I see '£1000 to £1234 per person per week' text on the page
+  When I click the top-level summary table Edit link for the section 'Pricing'
+  And I enter '100' in the 'Minimum price' field
+  And I click the 'Save and return to summary' button
+  Then I see '£100 to £1234 per person per week' text on the page
+
+Scenario Outline: Correct users cannot edit services
+  Given I am logged in as the production <role> user
+  When I am on the /admin/services/1123456789012346 page
```
TheDoubleK: Would be nice to do the random service thing here too I think, rather than relying on existing data.

PR #427, opened 2017-12-06
Update common steps to allow clicking links and buttons other than "Edit" in summary tables

benvand:
Update common steps to allow clicking links other than "Edit" in summary tables. Like 'Change name' or 'Update this'

Since we're already using 'Edit' in the step name we may as well pass this through as a variable. That way instead of just `Edit` the link can be anything.
See an example usage here, we want to click a 'Change name' link in a summary table:
https://github.com/alphagov/digitalmarketplace-functional-tests/pull/428/files#diff-0b5422a62fe3b478369c57195266b509R7

Also allow clicking buttons by `value` like in other steps


features/step_definitions/common_steps.rb
```diff
@@ -268,22 +268,26 @@
...
   edit_link.click
 end
 
-When /I click the summary table Edit link for '(.*)'$/ do |field_to_edit|
-  edit_link = page.find(:xpath, "//tr/*/span[contains(text(), '#{field_to_edit}')]/../..//a[text()]")
-  edit_link.text().should have_content('Edit')
+When /I click the summary table '(.*)' (link|button) for '(.*)'$/ do |link_name, elem_type, field_to_edit|
+  case elem_type
+    when 'link'
+      edit_link = page.find(:xpath, "//tr/*/span[text() = '#{field_to_edit}']/../..//a[contains(text(), '#{link_name}')]")
```
TheDoubleK: Does `text()` trim surrounding whitespace?  Off the top of my head I don't think it does, so I worry that moving from `contains` to equal that we're not going to be able to match properly if the `<span>` isn't closely packed around the text we're searching for.

features/step_definitions/common_steps.rb
```diff
@@ -268,22 +268,26 @@
...
+    else
+      edit_link = page.find(:xpath, "//tr/*/span[text() = '#{field_to_edit}']/../..//input[@value = '#{link_name}']")
```
TheDoubleK: Where do we have buttons at the top of a summary table?  That doesn't feel like a pattern we have anywhere?

benvand: Here, also the copy brief button IIRC

screen shot 2017-12-06 at 16 54 33

features/step_definitions/common_steps.rb
benvand: Let me check

benvand: Can indeed, Done 👍

PR #428, opened 2017-12-06
NEW: (Admin) Functional tests for update supplier name

benvand:
Requires 
 - [x] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/427


benvand:
@TheDoubleK Yeah that does sound better
I'm going to merge everything then change that everywhere as a single PR. We've got half in and half out at the moment

PR #429, opened 2017-12-06
NEW: (Admin) Functional tests for deactivating a supplier's contributor

benvand:
Requires 
 - [x] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/427

### DIFF
--- a/features/admin/deactivate_supplier_user.feature
+++ b/features/admin/deactivate_supplier_user.feature
@@ -1,0 +1,41 @@
+@deactivate-supplier-user
+Feature: Deactivate a supplier's contributor
+
+Scenario Outline: Correct users can deactivate and reactivate a supplier's contributor
+  Given I am logged in as the production <role> user
+  And I have a supplier with:
+    | name          | DM Functional Test Supplier |
+  And that supplier has a user with:
+    | name          | DM Functional Test Supplier User 1                               |
+    | email_address | dm-functional-test-supplier-user-1@digital.cabinet-office.gov.uk |
+  And that supplier has a user with:
+    | name          | DM Functional Test Supplier User 2                               |
+    | email_address | dm-functional-test-supplier-user-2@digital.cabinet-office.gov.uk |
+  And I click the 'Edit supplier accounts or view services' link
+  And I enter 'DM Functional Test Supplier' in the 'Find a supplier by name' field
+  And I click the 'find_supplier_by_name_search' button
+  And I click the summary table 'Users' link for 'DM Functional Test Supplier'
+  When I click the summary table 'Deactivate' button for 'DM Functional Test Supplier User 1'
+  Then I see an entry in the 'Users' table with:
+    | Name                               | Email address                                                    | Last login  | Pwd changed | Locked |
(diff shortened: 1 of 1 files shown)


features/admin/deactivate_supplier_user.feature
```diff
@@ -0,0 +1,26 @@
+@deactivate-supplier-user
+Feature: Deactivate a supplier's contributor
+
+Scenario Outline: Correct users can deactivate a supplier's contributor
+  Given I am logged in as the production <role> user
+  # TODO get here from /admin
```
TheDoubleK: Would be nice to fill the TODO steps to click through from the home page now admin links are finalised.

features/admin/deactivate_supplier_user.feature
```diff
@@ -0,0 +1,26 @@
...
+  And I am on the /admin/suppliers?supplier_name_prefix=DM+Functional+Test+Supplier page
+  And I click the summary table 'Users' link for 'DM Functional Test Supplier'
+  And I click the summary table 'Deactivate' button for 'DM Functional Test Supplier User 1'
```
TheDoubleK: This needs a check that the user is now actually deactivated.

features/admin/deactivate_supplier_user.feature
```diff
@@ -0,0 +1,26 @@
...
+  And I click the summary table 'Activate' button for 'DM Functional Test Supplier User 1'
```
TheDoubleK: And then a check that they have been re-activated.

features/admin/deactivate_supplier_user.feature
benvand: Done

features/admin/deactivate_supplier_user.feature
benvand: So this isn't actually possible in any normal kind of way (ie checking the page for new content). None of the values in that table change. However, the activate button is only present if they have already been deactivated. So finding and clicking activate kind of serves as the test. I've put in a step to check the table content doesn't change for that user when it's not supposed to though.

features/admin/deactivate_supplier_user.feature
benvand: As above

features/admin/deactivate_supplier_user.feature
```diff
@@ -0,0 +1,41 @@
+@deactivate-supplier-user
+Feature: Deactivate a supplier's contributor
+
+Scenario Outline: Correct users can deactivate a supplier's contributor
```
TheDoubleK: Really minor but should this be `Correct users can deactivate and reactivate a supplier's contributor`?

features/admin/deactivate_supplier_user.feature
TheDoubleK: I guess the way to do it would be to try to log in with that user account and check it's blocked.  But also the fact that the "Activate" button exists is probably enough.

features/admin/deactivate_supplier_user.feature
benvand: Done

PR #434, opened 2017-12-07
NEW: (Admin) Functional tests for finding a buyer

benvand:


features/admin/search_for_buyers.feature
```diff
@@ -0,0 +1,28 @@
...
+Feature: Find a buyer
+
+Scenario Outline: Correct users can search for buyers by brief id
+  Given I am logged in as the production <role> user
+  And I click the 'Find a buyer by opportunity ID' link
+  When I enter '5870' in the 'Find a buyer by opportunity ID' field
+  And I click the 'Search' button
+  Then I am on the 'Find a buyer' page
+  And I see the 'User' summary table filled with:
+    | Buyer Name 513ecc35174839416337590f218428e3 | 513ecc35174839416337590f218428e3@example.gov.uk |
```
TheDoubleK: Is this really the best account we have? Is it guaranteed to exist after data copying from prod?

benvand:
OK so it looks like 18723 will always be a buyer with email `18723@user.marketplace.team` and name `Test user` I think that one will be most stable. We'll have to see though.

PR #435, opened 2017-12-07
Opportunities search changes corresponding to new ES implementation

risicle:
This firstly fixes the tests to cover the changes corresponding to https://trello.com/c/KcqU7k20/101-suppliers-should-be-able-to-select-a-category-es-for-frontend and then extends them to exercise the new features.

### DIFF
--- a/features/smoke-tests/supplier/opportunities.feature
+++ b/features/smoke-tests/supplier/opportunities.feature
@@ -59,30 +59,85 @@ Scenario Outline: User can filter by individual location
     | International (outside the UK) |
     | Off-site                       |
 
-Scenario Outline: User can filter by both status and lot
+Scenario Outline: User gets no results for impossible combinations of location and lot
+  Given I am on the /digital-outcomes-and-specialists/opportunities page
+  And I click '<lot>'
+  And I check '<location>' checkbox
+  And I wait for the page to reload
+  Then I see no results
+  # now we backtrack a bit to check the lot "link" has been made unclickable when showing it has no results
+  When I click 'Clear filters'
+  And I wait for the page to reload
+  And I click 'All categories'
+  And I check '<location>' checkbox
+  And I wait for the page to reload
+  Then I don't see the '<lot>' link
+
+  Examples:
+    | lot                        | location                       |
+    | Digital specialists        | International (outside the UK) |
+    | Digital outcomes           | International (outside the UK) |
+    | User research participants | Off-site                       |
+
+Scenario Outline: User can filter by both status, lot and location together
   Given I am on the /digital-outcomes-and-specialists/opportunities page
   When I note the result_count
   And I click '<lot>'
+  Then I see that the stated number of results does not exceed that result_count
+  And I note the result_count
+  When I check '<status>' checkbox
   And I wait for the page to reload
   Then I see that the stated number of results does not exceed that result_count
   And I note the result_count
-  And I check '<status>' checkbox
-  And I wait for the page to reload
-  Then I see that the stated number of results does not exceed that result_count
   And I see all the opportunities on the page are on the '<lot>' lot
   And I see all the opportunities on the page are of the '<status>' status
+  When I check '<location>' checkbox
+  And I wait for the page to reload
+  Then I see that the stated number of results does not exceed that result_count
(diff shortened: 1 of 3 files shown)


risicle:
Oh and one test still fails due to the mismatch between `Off-site` and `Offsite` in the results. This is down to using the raw field value for result display instead of looking up the framework-defined label. @Wynndow is aware.

risicle:
I do have to wonder if some of these should be moved out of the `@smoke-tests`. The only reason they're in the smoke tests is they don't _happen to_ rely on mutating any state.

features/smoke-tests/supplier/opportunities.feature
```diff
@@ -34,46 +33,114 @@ Scenario Outline: User can filter by individual status
...
+  Given I am on the /digital-outcomes-and-specialists/opportunities page
+  And I click '<lot>'
+  And I check '<location>' checkbox
+  And I wait for the page to reload
+  Then I see no results
+  # now we backtrack a bit to check the lot "link" has been made unclickable when showing it has no results
+  When I click 'Clear filters'
+  And I wait for the page to reload
+  And I click 'All categories'
+  And I check '<location>' checkbox
```
galund: Other times after clicking on a filter, we've need a "wait for the page to reload" step. What's the cases where we do/don't need this?

risicle: In this particular case we probably should - I think  we might have just been getting lucky so far...

Generally I omit it when it's a link we're clicking.

features/smoke-tests/supplier/opportunities.feature
```diff
@@ -34,46 +33,115 @@ Scenario Outline: User can filter by individual status
...
   And I see all the opportunities on the page are of the '<status>' status
 
   Examples:
-    | status |
-    | Open   |
-    | Closed |
+    | status       |
+    | Open         |
+    | Closed       |
+    | Awarded      |
```
Wynndow: These are going to need to be removed. Just Open and Closed now.

features/step_definitions/opportunities_steps.rb
```diff
@@ -37,20 +40,41 @@
...
 
+Then (/^I see all the opportunities on the page are in the '(.*)' location/) do |lot|
+  locations_found = all(
+    :xpath,
+    '//*[@class="search-result"]//*[@class="search-result-important-metadata"][1]//*[@class="search-result-metadata-item"][2]'
+  )
+  locations_found.each { |x| x.text.should == lot }
+end
+
+Then (/^I see all the opportunities on the page are for the '(.*)' role/) do |lot|
```
Wynndow: Maybe change `|lot|` to `|role|`?

features/step_definitions/opportunities_steps.rb
```diff
@@ -37,20 +40,41 @@
...
+
 Then (/^I see all the opportunities on the page are of the '(.*)' status$/) do |status|
   published_or_closed = all(
     :xpath,
     '//*[@class="search-result"]//*[@class="search-result-metadata"][2]//*[@class="search-result-metadata-item"][1]'
   )
   published_or_closed.each do |x|
-    if status == 'Closed'
-      x.text.should == 'Closed'
+    if ['Closed', 'Unsuccessful', 'Cancelled'].include? status
```
Wynndow: Not a functional difference, but could remove `Unsuccessful` and `Cancelled`

features/smoke-tests/supplier/opportunities.feature
risicle: - [ ] Sigh.

features/step_definitions/opportunities_steps.rb
risicle: - [ ] Damnit

features/step_definitions/opportunities_steps.rb
risicle: Nah I'll leave it.

risicle:
Cool - need to think about the rollout of these I guess.

Wynndow:
Break the thing, fix the thing. That's the tradition isn't it?

risicle:
I've added stage-skipping tags.

PR #441, opened 2017-12-13
Remove aliased 'I have a buyer' and 'I have a supplier' user steps to make way for new supplier step

benvand:
We need a new `I have a supplier(? with name)` step to create an actual supplier object.
Like:
```
Given I have a supplier
 And that supplier has a user with name 'Test user 1'
 And that supplier has a user with name 'Test user 2'
 And that supplier has a user with name 'Test user 3'
```
These aliased names make the tests harder to understand because they don't describe what we're doing properly. A supplier needs to exist before a user object because a supplier can have multiple users.

The aliases also meant more digging through the code for what was actually going on.

### DIFF
--- a/features/step_definitions/login_steps.rb
+++ b/features/step_definitions/login_steps.rb
@@ -14,28 +14,6 @@ Given /^I have a production ([a-z-]+) user$/ do |user_role|
   puts "Email address: #{@user['emailAddress']}"
 end
 
-Given /^I have a ([a-z-]+) user(?: with supplier id (\d*))?$/ do |user_role, supplier_id|
-  randomString = SecureRandom.hex
-  password = ENV["DM_PRODUCTION_#{user_role.upcase.gsub('-', '_')}_USER_PASSWORD"]
-
-  user_data = {
-    "emailAddress" => randomString + '@example.gov.uk',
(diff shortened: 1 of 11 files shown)


benvand:
Ready and tested on staging

features/step_definitions/login_steps.rb
```diff
@@ -48,17 +26,17 @@
...
-  @buyer = step "I have a buyer user"
+Given 'I have a buyer user' do
+  @buyer_user = create_user('buyer')
 end
 
-Given 'I have a supplier' do
-  @supplier = create_supplier
-  @supplier_user = step "I have a supplier user with supplier id #{@supplier['id']}"
+Given 'I have a supplier user' do
+  @supplier = create_supplier()
```
TheDoubleK: What does adding the parentheses here change?  Nothing at all? Does Ruby execute the method even without the brackets?

benvand: It's nothing at all...
https://stackoverflow.com/a/342475 
My 🐍  sense is deeply unsettled.

benvand: Removed redundant parenthesis

PR #442, opened 2017-12-13
catalogue smoke tests: add test of no-results search term

risicle:
An app got pushed out which broke no-results pages of g9 search and the first we heard of it was production 500s appearing. This should make sure we catch that kind of thing earlier.

### DIFF
--- a/features/smoke-tests/buyer/catalogue.feature
+++ b/features/smoke-tests/buyer/catalogue.feature
@@ -102,3 +102,10 @@ Scenario: User is able to paginate through search results and all of the navigat
   When I click the Previous Page link
   Then I am taken to page 1 of results
   And I see the same number of category links as noted
+
+Scenario: User gets no results for an unfindable term
+  Given I am on the /g-cloud/search page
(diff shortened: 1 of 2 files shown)


features/smoke-tests/buyer/catalogue.feature
```diff
@@ -102,3 +102,10 @@ Scenario: User is able to paginate through search results and all of the navigat
   When I click the Previous Page link
   Then I am taken to page 1 of results
   And I see the same number of category links as noted
+
+Scenario: User gets no results for an unfindable term
+  Given I am on the /g-cloud/search page
+  And I enter 'metempsychosis' in the 'q' field
```
MuriloDalRi: What if a supplier starts offering metempsychosis as a service (MaaS)?

risicle: We'll all be turned into trees 🌳

MuriloDalRi: I like trees! 👍

PR #443, opened 2017-12-13
Change heading copy for brief responses application

laurenceberry:
Don't merge this PR before merging this PR: https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/16


katstevens:
Annoyingly we'll need two versions of each test, so that they'll pass on both preview and staging as we deploy.  

So something like:
```
# old version
@skip-local @skip-preview
When I enter 'example-email@gov.uk' in the 'respondToEmailAddress' field
   Then I am on the 'Check your answers' page
   ...

# new version
@skip-staging
When I enter 'example-email@gov.uk' in the 'respondToEmailAddress' field
   Then I am on the 'Check and submit your answers' page
   ...
```

PR #448, opened 2017-12-21
default.nix: phantomjs -> phantomjs2

risicle:
recent changes in cloudfrount's tls setup would appear to have broken
operation with phantomjs 1.9.8


risicle:
(incidentally, this raises our requirement to a NixOS >= 17.09 environment)

PR #449, opened 2017-12-21
use bilaterally normalize-space()'ed text when looking for links

risicle:
This fixes e.g. the catalogue smoke test when a service with a `serviceName`
containing a double-space is chosen. also include an `escape_xpath`
implementation cribbed from the interwebs which we should make more use of.

The fact that this normalization is needed has an interesting implication: capybara is operating on the post-html-normalized body text?


risicle:
html has rules about implicit whitespace normalization of "text". capybara is operating its' xpath-y-ness on a copy of the document after that operation has been performed.

PR #450, opened 2017-12-22
Update ruby version

MuriloDalRi:


risicle:
Hmmm I'm not sure this file my `default.nix` pays any attention to so I have no painless way of testing this on my end.

PR #453, opened 2018-01-05
 New "I see an entry in the something table with:" method

benvand:
* New "I see an entry in the something table with:" method
* Update *broken* feature to use new 'entry exists in table' method

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -336,6 +336,25 @@ Then /^I see that the '(.*)' summary table has (\d+)(?: or (more|fewer))? entr(?
   end
 end
 
+Then /^I see an entry in the '(.*)' table with:$/ do |table_heading, table|
+  expected_row = table.rows[0]
+  result_table_rows = get_table_rows_by_caption(table_heading)
+  match = false
+
+  result_table_rows.each do |row|
+    # Get the row as an array of strings to compare with our expected row
+    row_text_values = row.all('td').map {|td| td.text}
+    # Ensure that the expected strings are in the correct place and that we skip anything with '<ANY>'
+    if expected_row.each_with_index.map {|expected_value, expected_index| expected_value == '<ANY>' or row_text_values[expected_index] == expected_value}.all?
+      match = true
+      break
+    end
+  end
+
+  match.should be(true), "Expected: #{expected_row.join(' | ')}"
+end
+
+
 Then /^I see the closing date of the brief in the '(.*)' summary table$/ do |table_heading|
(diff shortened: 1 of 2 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -336,6 +336,24 @@
   end
 end
 
+Then /^I see an entry in the '(.*)' table with:$/ do |table_heading, table|
+  result_table_rows = get_table_rows_by_caption(table_heading)
+  expected_row = table.rows[0]
+  for row in result_table_rows
+    result_columns = row.all('td')
+    match = true
+    for value, index in expected_row.each_with_index
```
TheDoubleK: The nested loops and overwriting the "match" hurt my head a bit, but it's doing the right thing and I can't immediately suggest a neater way to do it...

Maybe a couple of comments would help, like "checking each field of a row", "reset if a row doesn't match", etc.

features/step_definitions/common_steps.rb
```diff
@@ -336,6 +336,24 @@
...
+      if match and value != '<ANY>' and value != result_columns[index].text
```
CrystalPea: if match is overwritten to true by us do we need to check for it in the if statement?

features/step_definitions/common_steps.rb
```diff
@@ -336,6 +336,24 @@
   end
 end
 
+Then /^I see an entry in the '(.*)' table with:$/ do |table_heading, table|
+  result_table_rows = get_table_rows_by_caption(table_heading)
+  expected_row = table.rows[0]
+  for row in result_table_rows
```
CrystalPea: You can also use 'each' loop, which is more idiomatic to Ruby, but then as we are a Python team, 'for' loop might be preferable for readability, your call really :).

https://stackoverflow.com/questions/3294509/for-vs-each-in-ruby
https://teamtreehouse.com/library/iteration-with-each

features/step_definitions/common_steps.rb
```diff
@@ -336,6 +336,25 @@
   end
 end
 
+Then /^I see an entry in the '(.*)' table with:$/ do |table_heading, table|
+  expected_row = table.rows[0]
+  result_table_rows = get_table_rows_by_caption(table_heading)
+  match = false
+
+  for row in result_table_rows
```
CrystalPea: You can also use 'each' loop, which is more idiomatic to Ruby, but then as we are a Python team, 'for' loop might be preferable for readability, your call really :).

https://stackoverflow.com/questions/3294509/for-vs-each-in-ruby
https://teamtreehouse.com/library/iteration-with-each

benvand:
@CrystalPea Realised what you meant this morning 😂  Fixed that

Thanks for the reviews @TheDoubleK @CrystalPea ! Merging

PR #458, opened 2018-01-08
Update supplier name

benvand:


features/step_definitions/common_steps.rb
```diff
@@ -120,18 +120,14 @@
 
 When /I click #{MAYBE_VAR} ?(button|link)?$/ do |button_link_name, elem_type|
   if elem_type == 'button'
-    page.click_button(button_link_name)
+    page.find(:xpath, "//input[@value='#{button_link_name}'] | //input[@name='#{button_link_name}'] | //button[text()='#{button_link_name}']").click
```
TheDoubleK: The scenario steps here start `I click the ...` so I'm not sure this will work?  i.e. it will be looking for a button with `the 'Save'` as it's text?

benvand: Maybe var is defined as a string here :( :
`digitalmarketplace-functional-tests/features/support/transforms.rb` Can't say I'm a massive fan of the pattern but I used it for consistency :/

TheDoubleK: Ah yeah I see, it'll strip the `the` out.

PR #459, opened 2018-01-08
Update deactivate-supplier-user to handle all roles

benvand:


features/admin/deactivate_supplier_user.feature
```diff
@@ -28,14 +30,31 @@ Scenario Outline: Correct users can deactivate and reactivate a supplier's contr
...
+  And I see an entry in the 'Users' table with:
+    | Name                               | Email address                                                    | Last login  | Pwd changed | Locked | Status |
+    | DM Functional Test Supplier User 2 | dm-functional-test-supplier-user-2@digital.cabinet-office.gov.uk | <ANY>       | <ANY>       | No     | Active |
+
+  Examples:
+    | role                    | link-name                   |
+    | admin-ccs-category      | Edit suppliers and services |
+    | admin-framework-manager | View suppliers and services |
+
+Scenario Outline: Correct users cannot view suppliers users
```
galund: I know it's not your naming, but "correct...cannot" is a very confusing way of explaining what's being tested here... I think I'd maybe write "users with insufficiently privileged roles cannot" or something.

benvand: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/428#issuecomment-355919388

benvand: Addressed elsewhere. Going to do it in another PR. This PR i to get the functional tests green again

features/admin/deactivate_supplier_user.feature
```diff
@@ -28,14 +30,31 @@ Scenario Outline: Correct users can deactivate and reactivate a supplier's contr
     | role  |
     | admin |
 
-Scenario Outline: Correct users cannot update suppliers users
+Scenario Outline: Correct users can view suppliers users
```
galund: Can this be "can view but not edit" or "can view but not deactivate"? i.e. something to qualify that we aren't just checking "visibility" here.

benvand: Done, Thanks for the review! :100:

PR #462, opened 2018-01-30
catalogue/opportunity search tests: also check presence and connectivity of aria-live regions

risicle:
Also take the chance to unify some functionality between service & brief search, add some useful verbosity to the execution.

See https://trello.com/c/qUViIfu9/172-dynamic-content-on-the-search-pages-which-updates-without-a-page-reload-is-marked-up-using-wai-aria-live-regions and https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/705

### DIFF
--- a/features/step_definitions/catalogue_steps.rb
+++ b/features/step_definitions/catalogue_steps.rb
@@ -47,18 +47,20 @@ Then(/^I see #{MAYBE_VAR} in the search summary text$/) do |value|
 end
 
 Then (/^I note the number of search results$/) do
-  @service_count = CatalogueHelpers.get_service_count(page)
+  @result_count = CatalogueHelpers.get_service_count(page)
+  puts "Noted result_count: #{@result_count}"
 end
 
 Then /^I click a random category link$/ do
   links = CatalogueHelpers.get_category_links(page)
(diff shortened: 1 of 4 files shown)


features/step_definitions/catalogue_steps.rb
```diff
@@ -47,18 +47,20 @@
...
   links = CatalogueHelpers.get_category_links(page)
   link_el = links.sample
   @category_name = link_el.text.sub(/ \([0-9]*\)/, "")  # remove service count
+  puts "Clicking '#{link_el.text}'"
   link_el.click
 end
 
 Then(/^I see fewer search results than noted$/) do
-  expect(CatalogueHelpers.get_service_count(page)).to be < @service_count
+  expect(CatalogueHelpers.get_service_count(page)).to be < @result_count
```
brenetic: `get_result_count` ?

risicle: Yeah I could have dug further to unify the two sets of steps, but I was really just interested in getting the variable name the same so that they could share *some* steps.

risicle:
(this obviously has to wait for the merge & release of the feature)

PR #463, opened 2018-02-01
Remove Companies House Number from supplier creation flow

CrystalPea:
This has been done as this number is optional and now we will
let suppliers add it on the 'company details' page.

Related pull request in supplier-frontend: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/782

Trello ticket:
https://trello.com/c/coCrPhFZ/25-remove-companies-house-page-from-sign-up-journey

### DIFF
--- a/features/supplier/supplier_creation.feature
+++ b/features/supplier/supplier_creation.feature
@@ -1,4 +1,81 @@
-@supplier-creation
+@supplier-creation @skip-preview
+Feature: Create new supplier account
+
+Background:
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Looks good but will need to skip-preview/skip-staging versions I think to keep things green as this gets released.

CrystalPea:
@TheDoubleK good point, I will add these tags and double check the process and merge stuff later in the afternoon then :)

PR #468, opened 2018-02-09
Add customisable content for the continue/save button

samuelhwilliams:
## Summary
We've updated the content for the button on the DUNS number page, but our currently functional tests only expect to see 'Continue'. Have added the ability to customise this button text per step.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -296,11 +296,12 @@ When /I click the summary table '(.*)' (link|button) for '(.*)'$/ do |link_name,
   edit_link.click
 end
 
-When /I update the value of '(.*)' to '(.*)' using the summary table '(.*)' link/ do |field_to_edit, new_value, link_name|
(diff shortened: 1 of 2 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -296,11 +296,12 @@
...
 end
 
-When /I update the value of '(.*)' to '(.*)' using the summary table '(.*)' link/ do |field_to_edit, new_value, link_name|
+When /I update the value of '(.*)' to '(.*)' using the summary table '(.*)' link(?: and the '(.*)' button)?/ do |field_to_edit, new_value, link_name, button_name|
   summary_page = current_url
+  button_name ||= "Continue"
 
   step "I click the summary table '#{link_name}' link for '#{field_to_edit}'"
-  step "I enter '#{new_value}' in the '#{field_to_edit}' field and click its associated 'Continue' button"
+  step "I enter '#{new_value}' in the '#{field_to_edit}' field and click its associated '#{button_name}' button"
```
galund: I guess I'd have expected that there was a way to just say "submit the form" and it would find any submit button. That would have been nice.

PR #471, opened 2018-02-27
Update supplier account test following changes to details page.

CrystalPea:
Connected to this pull request :https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/799

### DIFF
--- a/features/supplier/supplier_account.feature
+++ b/features/supplier/supplier_account.feature
@@ -6,6 +6,7 @@ Background:
   And that supplier is logged in
   And I am on the /suppliers page
 
+@skip-preview
 Scenario: Supplier user can edit supplier information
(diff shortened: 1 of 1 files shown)


features/supplier/supplier_account.feature
```diff
@@ -28,6 +29,25 @@ Scenario: Supplier user can edit supplier information
     | Supplier summary          | All fresh                       |
 
 
+@skip-preview @skip-staging
```
TheDoubleK: This one should only be skip-staging I think.

CrystalPea: it needs to be @skip-preview first, so when I merge it, the tests won't fail with old version of the code. Then I merge new code, I merge new tests without @skip-preview tag

PR #474, opened 2018-03-02
Opportunities keyword search

risicle:
https://trello.com/c/qdNbRJ4e/331-supplier-can-search-for-buyer-on-dos

Added simple test for smoke tests, expanded into much more arduous tests for full suite which required me to fix up a few of the step definitions beforehand (who knew we were never able to uncheck a checkbox before?)

### DIFF
--- a/features/step_definitions/opportunities_steps.rb
+++ b/features/step_definitions/opportunities_steps.rb
@@ -20,16 +20,17 @@ Then (/^I see an opportunity in the search results$/) do
   page.should have_selector(:css, ".search-result")
 end
 
-Then (/^I see that the stated number of results does not exceed that (\w+)$/) do |variable_name|
-  var = instance_variable_get("@#{variable_name}")
+Then (/^I see that the stated number of results (does not exceed|equals|is no fewer than) that (\w+)$/) do |comparison_string, variable_name|
+  var_val = instance_variable_get("@#{variable_name}")
   @new_result_count = page.first(:css, ".search-summary-count").text.to_i
   puts "Number of results: #{@new_result_count}"
-  @new_result_count.should <= var
-end
-
-Then (/^I see that the stated number of results equals that (\w+)$/) do |variable_name|
-  var = instance_variable_get("@#{variable_name}")
-  var.should == page.first(:css, ".search-summary-count").text.to_i
+  if comparison_string == "does not exceed"
+    @new_result_count.should <= var_val
+  elsif comparison_string == "equals"
+    @new_result_count.should == var_val
+  elsif comparison_string == "is no fewer than"
+    @new_result_count.should >= var_val
+  end
 end
 
 Then (/^I see all the opportunities on the page are on the '(.*)' lot$/) do |lot|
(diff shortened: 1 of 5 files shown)


features/step_definitions/opportunities_steps.rb
```diff
@@ -20,16 +20,17 @@
...
-end
-
-Then (/^I see that the stated number of results equals that (\w+)$/) do |variable_name|
-  var = instance_variable_get("@#{variable_name}")
-  var.should == page.first(:css, ".search-summary-count").text.to_i
+  if comparison_string == "does not exceed"
+    @new_result_count.should <= var_val
+  elsif comparison_string == "equals"
+    @new_result_count.should == var_val
+  elsif comparison_string == "is no fewer than"
```
galund: I'd have preferred "is at least" but you might want to canvass others to see what's more readable. I mean, I find `>=` more readable than text altogether, but I suppose that's not really the idea :-)

features/supplier/opportunities.feature
```diff
@@ -102,18 +102,43 @@ Scenario Outline: User can filter by both status, lot and location together
...
+    | User research participants | Open         | Off-site                       | services            |
+    | Digital specialists        | Closed       | South West England             | enterprise software |
+    | Digital outcomes           | Closed       | International (outside the UK) | delivery            |
+    | User research participants | Closed       | Off-site                       | "agile methodology" |
+    | Digital specialists        | Open         | London                         | improve             |
+    | Digital outcomes           | Closed       | International (outside the UK) | performance         |
+    | User research participants | Open         | Off-site                       | analyze             |
+    | Digital specialists        | Closed       | Wales                          | management          |
+    | Digital outcomes           | Open         | Yorkshire and the Humber       | national            |
+    | User research participants | Closed       | Northern Ireland               | metempsychosis      |
```
galund: That we have to list all of these out in our tests feels like technical debt of some sort. If you agree then please raise a ticket, I'm not saying we should tackle that here :-) We avoided it for G-Cloud, anyway.

features/support/helpers.rb
```diff
@@ -64,7 +64,7 @@ def check_checkbox(locator_or_checkbox, options={})
 def uncheck_checkbox(locator_or_checkbox, options={})
   checkbox = return_element('checkbox', locator_or_checkbox, options)
 
-  check(checkbox[:id], options.merge({allow_label_click: true}))
+  uncheck(checkbox[:id], options.merge({allow_label_click: true}))
```
galund: :-D

idavidmcdonald:
I think you are going to run into the fun `skip-staging` stuff needed as these will fail on non preview environments :(

features/step_definitions/opportunities_steps.rb
risicle: I agree if you look at it on its own but if you look at it in the sentence, it doesn't seem to fit as well, especially given that the text following it can be any valid `MAYBE_VAR` entry...

features/supplier/opportunities.feature
risicle: Well, we're just picking out a smattering of combinations to test. It would be infeasible to test all combinations and I think it would be pretty tricky to do that in cucumber. Got to remember that cucumber is pretty limited - my `MAYBE_VAR` stuff alone was a hack pushing at its limits.

risicle:
@idavidmcdonald I wasn't particularly planning on merging this before the feature had made it to production - there seemed a lot of keenness to just get the damn feature out at the end of last week - not sure where that went.

PR #475, opened 2018-03-08
Update DOS search results output

brenetic:
We now output more detail in the DOS search results with regard to the
status of an award.

See alphagov/digitalmarketplace-buyer-frontend#723

Also added `pry` gem for debugging.

https://trello.com/c/y59qiKOl/335-show-award-status-in-dos-search-results

### DIFF
--- a/features/step_definitions/opportunities_steps.rb
+++ b/features/step_definitions/opportunities_steps.rb
@@ -63,8 +63,8 @@ Then (/^I see all the opportunities on the page are of the '(.*)' status$/) do |
     '//*[@class="search-result"]//*[@class="search-result-metadata"][2]//*[@class="search-result-metadata-item"][1]'
   )
   published_or_closed.each do |x|
-    if ['Closed', 'Unsuccessful', 'Cancelled'].include? status
-      x.text.should == status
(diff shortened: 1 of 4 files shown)


features/step_definitions/opportunities_steps.rb
```diff
@@ -63,8 +63,8 @@
     '//*[@class="search-result"]//*[@class="search-result-metadata"][2]//*[@class="search-result-metadata-item"][1]'
   )
   published_or_closed.each do |x|
-    if ['Closed', 'Unsuccessful', 'Cancelled'].include? status
-      x.text.should == status
+    if %w[Closed Unsuccessful Cancelled].include? status
```
MuriloDalRi: This array syntax might be a little confusing for people not used to Ruby.

katstevens: Ooo that looks weird 😹

katstevens: Do we need `Awarded` as well?

brenetic: Literal arrays are used by the Ruby community generally instead of the more traditional method. Will swap it back and also add the awarded.

MuriloDalRi:
LGTM 👍

PR #476, opened 2018-03-08
Skip staging for DOS search result changes

brenetic:
This should have been part of https://github.com/alphagov/digitalmarketplace-functional-tests/pull/475


TheDoubleK:
But we don't have any version of those tests for staging now?  Generally you'd want a `skip-preview` version of the test that works for staging still, so we maintain test coverage of the features.

brenetic:
@TheDoubleK Have updated with what I think will work?

PR #480, opened 2018-03-14
Update DOS search results steps to show award status details

brenetic:
We now output more detail in the DOS search results with regard to the status of an award.

This PR re-reverts alphagov/digitalmarketplace-functional-tests#477

Also refactor the opportunity steps a little bit.

Related to: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/732

https://trello.com/c/y59qiKOl/335-show-award-status-in-dos-search-results

### DIFF
--- a/features/support/helpers.rb
+++ b/features/support/helpers.rb
@@ -11,7 +11,14 @@ LOTS = {
   PaaS: 'Platform as a Service',
   IaaS: 'Infrastructure as a Service',
   SCS: 'Specialist Cloud Services'
-}
+}.freeze
+
+META_DATA_XPATHS = {
+  location: '//*[@class="search-result"]//*[@class="search-result-important-metadata"][1]//*[@class="search-result-metadata-item"][2]',
+  lot: '//*[@class="search-result"]//*[@class="search-result-metadata"][1]//*[@class="search-result-metadata-item"][1]',
+  role: '//*[@class="search-result"]//*[@class="search-result-metadata"][1]//*[@class="search-result-metadata-item"][2]',
+  status: '//*[@class="search-result"]//*[@class="search-result-metadata"][2]//*[@class="search-result-metadata-item"][1]'
+}.freeze
 
 CLOSED_OUTCOMES = {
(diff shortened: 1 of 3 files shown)


features/support/helpers.rb
```diff
@@ -11,12 +11,38 @@
...
+  role: '//*[@class="search-result"]//*[@class="search-result-metadata"][1]//*[@class="search-result-metadata-item"][2]',
+  status: '//*[@class="search-result"]//*[@class="search-result-metadata"][2]//*[@class="search-result-metadata-item"][1]'
+}.freeze
+
+CLOSED_OUTCOMES = {
+  Awarded: 'Closed: awarded',
+  Cancelled: 'Closed: cancelled',
+  Closed: 'Closed: awaiting outcome',
+  Unsuccessful: 'Closed: no suitable suppliers'
+}.freeze
```
katstevens: Guessing this makes `CLOSED_OUTCOMES` unmutable?

features/support/helpers.rb
```diff
@@ -11,12 +11,38 @@
...
 def full_lot(lot)
   LOTS[lot.to_sym]
 end
 
+def search_result_metadata_items(type)
+  all(:xpath, META_DATA_XPATHS[type])
+end
+
+def closed_outcome?(status)
+  CLOSED_OUTCOMES.key?(status.to_sym)
```
katstevens: I am going to have to properly read up on Ruby symbols at some point, aren't I... 😸

features/support/helpers.rb
brenetic: Yes indeed :)

features/support/helpers.rb
brenetic: Yeah, the only reason I did that was because there was the `LOTS` dictionary making use of symbols as keys. Otherwise a string key hash would have sufficed without the need to convert the passed in key to symbol:

```
hash = {
  'foo' => 'bar
}
```

Ruby loves to give us option! 😃

PR #482, opened 2018-03-14
Tests 142 search clear filters behaviour with keyword search

benvand:


samuelhwilliams:
Nice one. Do we also want to add a specific test to check that the clear filters link points to the correct search terms (i.e the text in the search box is the same before/after clicking the link)?

benvand:
@samuelhwilliams Done 😡 😩 😭 I had to learn xpath syntax!

risicle:
@benvand xpath can be fairly awesome, especially on the python side with `lxml` and the extra tricks it provides.

PR #486, opened 2018-03-16
Use RSpec `expect` syntax

brenetic:
This swaps all the `should` syntax to use the newer `expect` syntax. The RSpec blog has a good explanation for their reasoning: http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/


samuelhwilliams:
Nice link to docs explaining the change. Super helpful. I don't feel overly confident critiquing this PR, but it looks good to me from the look over I've given it. Going to try to get them running locally though...

PR #492, opened 2018-03-20
Add govuk linter

brenetic:
This adds the govuk-lint gem which is based on Rubocop. It is admittedly quite harsh. To start with only 2 of the rules have been enabled to give an example. It is best to look at each commit instead of the entire change.

The purpose of this is: 

- To bring this more inline with how Ruby is written across GDS. 
- To provide examples of some common Rubyisms for those not familiar with the language.

The config file added has a lot of the linting turned off at this point. Over time they will be enabled with the corrections made. 

The `run` task in the `Makefile` now runs the linter as standard but will only look for changes from this time going forward. This may or may not be a good idea at this time.


samuelhwilliams:
Hey this looks good.

TheDoubleK:
Yeah seems like it would be good to inject a bit of rigour to our Ruby hacks. I suggest you merge it sometime next Thursday, once I'm out of the picture.

brenetic:
Would people like the linter run when CI runs the functional tests?

katstevens:
It'd be good to have Travis run it as a pre-merge check - not sure if it would add value during the test run itself?

PR #498, opened 2018-03-26
Set default notify key

samuelhwilliams:
## Summary
If DM_NOTIFY_API_KEY has not been defined, automatically decrypt
credentials (from DM_CREDENTIALS_REPO) and inject it. This should make
it easier to run tests locally - though you will still need to have
started your apps with this environment variable in place. Also adds the
`@notify` tag to two scenarios which were missing it.

This is kinda a strawman as it does mean that credentials will get decrypted every time (requiring 2fa input) even if you specify only a subset of the tests. Will this be faff/annoying?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,10 @@ run-parallel: setup
 build-report:
 	report_builder -s 'reports' -o 'reports/index' -f html -t features,errors -T '<img src="https://media.giphy.com/media/sIIhZliB2McAo/giphy.gif">Functional Test Results'
 
-setup: install clean
+env:
+  DM_NOTIFY_API_KEY := $(if ${DM_NOTIFY_API_KEY}, ${DM_NOTIFY_API_KEY}, $(shell ${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/vars/preview.yaml | grep notify_api_key | sed 's/^.* //'))
+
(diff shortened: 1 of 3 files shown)


katstevens:
I can see this being frustrating e.g. if someone has left their phone at home for the day.

Perhaps we could have a setting in `local.sh`/`example.sh` that would override this and skip the `@notify` tests if `DM_NOTIFY_API_KEY` is not present?

samuelhwilliams:
Hmm, well I guess a simply override for this would be to set `DM_NOTIFY_API_KEY` in `local.sh` to any value. If the env var is set, it won't try to decrypt credentials.

katstevens:
I guess it boils down to whether new/phoneless devs can live with a couple of tests failing noisily (this is only mildly annoying at the moment).

samuelhwilliams:
I feel like it's probably worse to have a thing magically skip tests than for them to fail because they're not set up correctly.

samuelhwilliams:
Any input @galund?

galund:
I guess I'm concerned about the "you will still need to have started your apps with this environment variable in place" thing: is it common to have the env var in place when starting up an app, but then not have it when running tests?

It might be worth doing a thought experiment about an optimal way to configure an environment for  the frontend apps too, is maybe all I'm saying. Because that might share something in common with what we do for the functional tests.

PR #501, opened 2018-03-27
Remove doc_type from search result params

brenetic:
We no longer have a hidden field for `doc_type` for use with analytics. We are able to use the 'Destination Page' dimension against the 'Search Term' dimension filtering by either '/g-cloud/' or '/digital-outcomes-and-specialists/' which provides more reliable data.

Merge after:
- [x] https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/738

https://trello.com/c/8RzdZPsZ/393-buyer-frontend-search-page-doctype-parameter-is-broken-and-should-be-removed

### DIFF
--- a/features/smoke-tests/public/live_reload.feature
+++ b/features/smoke-tests/public/live_reload.feature
@@ -1,6 +1,30 @@
 @live-reload
 Feature: Live reload search
 
+@skip-staging
+Scenario: Live reload search hitting enter should not cause a full page reload
+  Given I am on the homepage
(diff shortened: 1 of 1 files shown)


risicle:
I'm not quite sure how your change correlates with your description here...

risicle:
Ok. Explained out of band. The tests are being temporarily cloned, one copy targeted at staging with `doc_type` remaining, one targeted at preview with `doc_type` gone.

PR #503, opened 2018-03-29
Logical operator precedence

brenetic:
Ruby's `and` and `or` are not the same as Python's.

```
a = true and false
a => true

a = true && false
a => false
```

This is because of operator [precedence](https://ruby-doc.org/core-2.3.1/doc/syntax/precedence_rdoc.html).

You might see `and` and `or` used like so:

`do_something_awesome or raise RuntimeError "All the awesome is gone"`

Though quite often the above is written as:

`raise RuntimeError "All the awesome is gone" unless do_something_awesome`

`then` is also usually not used in conditional statements.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,16 +12,10 @@ EmptyLineBetweenDefs:
 TrailingBlankLines:
   Enabled: false
 
-AndOr:
-  Enabled: false
-
 MethodCalledOnDoEndBlock:
   Enabled: false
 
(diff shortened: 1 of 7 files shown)


brenetic:
To follow up, the reason why the tests continued to pass while using `and` and `or` is because coincidentally the lower precedence did not affect their use in those particular boolean expressions.

Most will just use `&&` and `||` to avoid accidentally adding a subtle, unintended bug.

PR #505, opened 2018-04-04
User research tests

benvand:
Tests for signing up to user research and asserting that banner is only visible at appropriate times

* Expand radio button checks to also be used with checkbox
* Rename login_steps to user_steps
* Add step to sign user up to user research mailing list
* Add checks for user research banner on buyer account page
* Add checks for user research banner on supplier account page
* Add tests for user research sign up for suppliers and buyers


features/step_definitions/user_steps.rb
```diff
@@ -90,3 +90,13 @@
     And I don't see the 'Log out' button
   }
 end
+
+Given /^that user is on the user research mailing list$/ do
+  steps %Q{
```
brenetic: `Q` is the default so can leave it out

benvand:
Thanks @brenetic, I've made those changes

brenetic:
👍 coolio!

PR #506, opened 2018-04-04
standardize email addresses used across functional tests

risicle:
include section in README.md listing and explaining the vetted domains

Trello https://trello.com/c/YQB2HkiD/225-functional-tests-send-emails-to-invalid-email-addresses

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,3 +153,25 @@ Setting up your local environment and database to run the functional tests again
   - Building your app locally will create a local image of it and give it the `:latest` tag.
   - Docker Compose will now use your local copy including your new stuff.
   - Then start the local container environment as described above.
+
+## Principles to follow when writing functional tests
+
+### Email addresses
+
+When using "dummy" email addresses in functional tests, stick to emails using one of the following domains:
+
+| Email domain            | Useful properties                                                                       |
+|-------------------------|-----------------------------------------------------------------------------------------|
+| `example.com`           | Buyer-account-invalid                                                                   |
+| `example.gov.uk`        | Buyer-account-valid                                                                     |
+| `user.marketplace.team` | Admin-account-valid, can run tests against existing users with existing services/briefs in the database, doesn't look immediately "fake" to external services |
+
+These domains have been added to the frontends' `DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS` config setting and should have
+any Notify emails destined for them redirected to a "safe" dummy address lest Notify lose karma with their downstream
+provider from all the nonexistent-address bounces. If a different domain *is* required, it should probably be added here
+and synchronized across all the frontends' settings.
+
+Even if the test being written isn't expected to result in a Notify message (e.g. maybe an address only expected to be
+used against Mailchimp or just stored in a declaration) it's still best to stick to this vetted list of email domains -
+you never know when someone's going to quietly add a new notification of an event, and it's quite useful to have a
+handle on what email addresses we're throwing about in general in case we get any other complaints from other services.
--- a/features/admin/invite_admin_user.feature
+++ b/features/admin/invite_admin_user.feature
@@ -5,10 +5,10 @@ Scenario Outline: Admin Manager user can log in and invite admin users
   Given I am logged in as the production <role> user
   And I click the 'View and edit admin accounts' link
   And I click the 'Invite user' link
-  When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
+  When I enter 'some-desired-admin@user.marketplace.team' in the 'Email address' field
   And I choose the 'Manage services' radio button
   And I click the 'Invite user' button
-  Then I see a success banner message containing 'An invitation has been sent to simulate-delivered@notifications.service.gov.uk.'
+  Then I see a success banner message containing 'An invitation has been sent to some-desired-admin@user.marketplace.team.'
 
   Examples:
     | role          |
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -153,3 +153,24 @@ Setting up your local environment and database to run the functional tests again
   - Building your app locally will create a local image of it and give it the `:latest` tag.
   - Docker Compose will now use your local copy including your new stuff.
   - Then start the local container environment as described above.
+
+## Principles to follow when writing functional tests
+
+### Email addresses
+
+When using "dummy" email addresses in functional tests, try to stick to emails using one of the following domains:
```
katstevens: Do or do not - there is no `try`!

README.md
```diff
@@ -153,3 +153,24 @@ Setting up your local environment and database to run the functional tests again
...
+| `user.marketplace.team` | Admin-account-valid, can run tests against existing users with existing services/briefs in the database |
+
+These domains have been added to the frontends' `DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS` config setting and should have
+any Notify emails destined for them redirected to a "safe" dummy address lest Notify lose karma with their downstream
+provider from all the nonexistent-address bounces. If a different domain *is* required, it should probably be added here
+and synchronized across all the frontends' settings.
+
+Even if the test being written isn't expected to result in a Notify message it's still best to stick to this vetted list
+of email domains - you never know when someone's going to add a new notification of an event, and it's quite useful
+to have a handle on what email addresses we're throwing about in general in case we get any other complaints.
```
katstevens: Worth mentioning that Mandrill/Mailchimp isn't affected (or is it? Is there a separate whitelist address?)

features/supplier/subscribe_mailing_list.feature
```diff
@@ -13,10 +13,10 @@ Background:
   And I don't see a banner message
 
 Scenario: Successful mailing-list subscription from the home page
-  When I enter 'functional-test-example-email@gov.uk' in the 'email_address' field
+  When I enter 'functional-test-email@user.marketplace.team' in the 'email_address' field
   And I click 'Subscribe'
   Then I am on the 'Digital Marketplace' page
-  And I see a success banner message containing 'You will receive email notifications to functional-test-example-email@gov.uk when applications are opening.'
+  And I see a success banner message containing 'You will receive email notifications to functional-test-email@user.marketplace.team when applications are opening.'
```
katstevens: As per previous comment, this one is Mailchimp rather than Notify, right? Can we use `user.marketplace.team` in the same way here?

features/admin/invite_admin_user.feature
```diff
@@ -5,10 +5,10 @@ Scenario Outline: Admin Manager user can log in and invite admin users
   Given I am logged in as the production <role> user
   And I click the 'View and edit admin accounts' link
   And I click the 'Invite user' link
-  When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
+  When I enter 'some-desired-admin@user.marketplace.team' in the 'Email address' field
```
katstevens: This makes it sound like the admin is Helen of Troy...

README.md
risicle: I'm going to pretend not to get that reference...

README.md
risicle: Yup probably worth being explicit.

features/supplier/subscribe_mailing_list.feature
risicle: Well, as far as I'm aware, `@user.marketplace.team` isn't hooked up to anything in exactly the same way `@gov.uk` isn't, so I imagine it will have the same effect.

risicle: Well, as far as I'm aware, `@user.marketplace.team` isn't hooked up to anything in exactly the same way `@gov.uk` isn't, so I imagine it will have the same effect.

features/admin/invite_admin_user.feature
risicle: I see what you did there.

README.md
katstevens: Translation: avoiding ambiguity in docs is a good thing 😸

features/supplier/subscribe_mailing_list.feature
katstevens: I double checked the manual and it says that *any* Mandrill email on preview/staging goes to a sandbox (is that right?) https://alphagov.github.io/digitalmarketplace-manual/emails.html?highlight=emails#id3

katstevens: But Mailchimp doesn't have a sandbox - there are preview/staging mailing lists instead.

README.md
risicle: Well, ijust... don't want to be too much of a ball-breaker and realize that I may not have considered every circumstance with the 3 I've provided here.

katstevens: Fair enough!

features/supplier/subscribe_mailing_list.feature
risicle: I ... don't know. Didn't think we still used mandrill itself? Still, this test works with `user.marketplace.team`, and changing it reduces the number of domains we use.

README.md
risicle: (because I'm likely to be the first to break the rules and am foreseeing my hypocrisy)

samuelhwilliams: I think @katstevens is right. It's good to be specific in our documentation. If we discover a new and justified thing that breaks the rules, then we can always update the docs. ;)

risicle: changed

PR #509, opened 2018-04-12
Add travis config file

brenetic:
This will only run the govuk ruby linter at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: ruby
+rvm:
+  - 2.3.1
+install:
+  - gem install govuk-lint
+script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: ruby
+rvm:
+  - 2.3.1
+install:
+  - bundle install --without development test
```
katstevens: In the `Makefile` we have `bundle install --path .bundle --without development test`. I take it we don't need the `--path` arg here?

katstevens: That sounds good to me!

PR #511, opened 2018-04-23
Fix typo in api helpers

brenetic:
~~expet~~ => expect


brenetic:
It gets ignored unfortunately, which is surprising as I thought `expect` was a method call . . .

PR #512, opened 2018-04-26
Set supplier org size on supplier instead of declaration

Wynndow:
We now use the org size from the supplier object when serializing
brief-responses. If a supplier doesn't have their org size set (which
they have to to be able to apply to a framework), we see errors in the
frontend.

This update moves setting the org size from the declaration to the supplier.

screen shot 2018-04-26 at 18 21 44

### DIFF
--- a/features/step_definitions/supplier_steps.rb
+++ b/features/step_definitions/supplier_steps.rb
@@ -36,7 +36,8 @@ end
 
 Given /^that(?: (micro|small|medium|large))? supplier has applied to be on that framework$/ do |organisation_size|
   organisation_size ||= %w[micro small medium large].sample
-  submit_supplier_declaration(@framework['slug'], @supplier["id"], 'status': 'complete', 'organisationSize': organisation_size, 'nameOfOrganisation': 'foobarbaz', 'primaryContactEmail': 'foo.bar@example.com')
+  update_supplier(@supplier["id"], 'organisationSize': organisation_size)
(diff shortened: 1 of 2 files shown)


features/step_definitions/supplier_steps.rb
```diff
@@ -36,7 +36,8 @@
 
 Given /^that(?: (micro|small|medium|large))? supplier has applied to be on that framework$/ do |organisation_size|
   organisation_size ||= %w[micro small medium large].sample
-  submit_supplier_declaration(@framework['slug'], @supplier["id"], 'status': 'complete', 'organisationSize': organisation_size, 'nameOfOrganisation': 'foobarbaz', 'primaryContactEmail': 'foo.bar@example.com')
+  update_supplier(@supplier["id"], 'organisationSize': organisation_size)
+  submit_supplier_declaration(@framework['slug'], @supplier["id"], 'status': 'complete', 'nameOfOrganisation': 'foobarbaz', 'primaryContactEmail': 'foo.bar@example.com')
```
katstevens: Interesting - did we get rid of the org size in the declaration entirely, then?

Wynndow: No, we still have it in the declaration I think. We make sure it's there when we set a framework status to `pending` (https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/frameworks.py#L152). This fix is needed because we serialized brief responses with the org size stored on the supplier and not their declaration now (https://github.com/alphagov/digitalmarketplace-api/pull/775).

PR #517, opened 2018-05-04
Update reset password flow to contain new field names

CrystalPea:
Following changes to user-frontend: https://github.com/alphagov/digitalmarketplace-user-frontend/pull/58/files


samuelhwilliams:
Will need to be careful with this as it will break functional tests on staging until user gets shipped out. But given its Friday and nothing else should ship, it is probably OK to leave until next week.

CrystalPea:
@samuelhwilliams good point 👍

samuelhwilliams:
Nope ignore my point! It was a bad point. The test you've written is already `skip-staging` :)

PR #520, opened 2018-05-11
Add line showing how to use rbenv

lfdebrux:
It took me a little bit of digging to figure out how to use `rbenv`, so I wanted to add it to the manual to make life easier in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,16 @@ Technology wise it is a Ruby project using:
 
 ## Bootstrapping the project on Mac
 
-This installation assumes you're using [Homebrew](http://brew.sh/) and Ruby
-installed by Homebrew. We expect you to have [rbenv](https://github.com/rbenv/rbenv)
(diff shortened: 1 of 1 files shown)


brenetic:
`rbenv rehash` should be run automatically when a gem is installed, so technically we shouldn't need to run it manually.

We should be specific about the ruby version since we are explicitly stating it in our .ruby_version file

`rbenv install 2.3.1`

Now that I've read the install bit through I realise it is a bit misleading. If we're telling the user to use rbenv we shouldn't tell them to use Homebrew to install ruby.

Let's have a chat. If you had to use this README to install ruby from scratch it would be cool to get more of your feedback.

lfdebrux:
@brenetic I've updated the installation section based on your comments.

I also removed the part about issues with Nokogiri with native Mac Ruby, as if we're telling people to use rbenv it doesn't really seem relevant.

PR #524, opened 2018-05-21
Namespace supplier name

Wynndow:
We were having issues with supplier names clashing when running tests in
parallel. One of the methods we use for creating suppliers does a
`get_or_create` and uses a supplier name as a prefix for the `get`. By
having very similar names we could get into a situation that a required
supplier would never be created, if a similar supplier was created
first.

### DIFF
--- a/features/admin/invite_supplier_contributor.feature
+++ b/features/admin/invite_supplier_contributor.feature
@@ -4,10 +4,10 @@ Feature: Invite a contributor to a supplier account
 Scenario Outline: Correct users can invite a contributors to a supplier account
   Given I am logged in as the production <role> user
   And I have a supplier with:
-    | name          | DM Functional Test Supplier 1 |
+    | name          | DM Functional Test Supplier - Invite a contributor feature |
   And I click the 'Edit supplier accounts or view services' link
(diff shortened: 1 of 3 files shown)


features/admin/invite_supplier_contributor.feature
```diff
@@ -4,10 +4,10 @@ Feature: Invite a contributor to a supplier account
 Scenario Outline: Correct users can invite a contributors to a supplier account
   Given I am logged in as the production <role> user
   And I have a supplier with:
-    | name          | DM Functional Test Supplier 1 |
+    | name          | Invite a contributor feature DM Functional Test Supplier |
```
katstevens: I think we should do this for the other two admin supplier scenarios (`deactivate_supplier_user` and `update_supplier_name`) as well, to avoid future tests succumbing to the same problem.

I'd prefer to keep `DM Functional Test..` at the beginning of the string in case we do need to easily identify all the test accounts - would `DM Functional Test Invite A Contributor Supplier` still work?

Wynndow: That's a smashing point. I'll update.

Wynndow:
Have pushed up changes so all admin feature suppliers won't clash.

PR #525, opened 2018-05-28
Use unicode quotes rather than apostrophes

samuelhwilliams:
## Summary
We generally prefer unicode right quotes (u+2019) over apostrophes. I've
updated this apostrophe in the briefs-frontend but we have functional
tests asserting this content, so this needs updating as well.

https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/109

### DIFF
--- a/features/buyer/award_flow_requirements.feature
+++ b/features/buyer/award_flow_requirements.feature
@@ -40,7 +40,7 @@ Scenario: Award flow - Award a requirement to a winning supplier
   And I see a temporary-message banner message containing 'Value: £20,000'
   And I see a temporary-message banner message containing 'Company size'
 
-
+@skip-preview
 Scenario: Award Flow - Cancel requirements
 
   Given I choose 'No' radio button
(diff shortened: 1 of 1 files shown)


features/buyer/award_flow_requirements.feature
```diff
@@ -74,6 +91,23 @@ Scenario: Award flow - Mark requirements as unsuccessful
...
+  Then I am on the 'Why didn’t you award a contract for %s?' page with brief 'title'
+
+  When I choose 'There were no suitable suppliers' radio button
+  And I click the 'Update requirements' button
+  Then I am on the '%s' page with brief 'title'
+  And I see a success banner message containing 'updated'
+
+  When I go to that brief page
+  Then I see a temporary-message banner message containing 'No suitable suppliers applied'
+  And I see a temporary-message banner message containing 'The buyer didn't award this contract because no suppliers met their '
```
katstevens: Just to confirm, this message appears on the Buyer FE, right? So no apostrophe change needed yet for `The buyer didn't`?

samuelhwilliams: Yep!

PR #530, opened 2018-06-06
Make test go fast zoom zoom

samuelhwilliams:
## Summary
🚗 ⏩ 💨 💨 🚗 🏎 
Disable waiting for the majority of our calls to Capybara's matchers.
Our site is primarily static content after the page has loaded, so we
don't need Capybara to be willing to wait for theoretical ajax/dynamic
content to finish loading.

Where we do have dynamic content (such as live search on
services/briefs), we explicitly add a maximum wait time that Capybara
will respect.

Capybara's matchers/selectors work in a specific way to support dynamic
changes on the page:

When we ask Capybara to check for an element or some text on the page,
if the text is present, Capybara immediately notices this and returns
the result. If the text is *not* present, Capybara will wait for its
`max_default_wait_time` (seconds) configuration value, checking the page
every 0.05s to see if the content has appeared yet. After the wait time
has elapsed, if the content is still not present, Capybara raises an
'element not found' exception to report that the selector is missing.
This is all well and good if we expect the content to be there. An issue
arises when we use a 'look for this content' selector, and check that
the return value is `false`, to assert that content is *not* present. In
this case, we will always force Capybara to sleep out its
'max_default_wait_time' period. Where we want to do this, we should be
careful to use the negated matcher that Capybara often exposes (e.g.
there is a `has_content?` matcher for checking that content is present,
and a `has_no_content?` matcher for checking that content is *not*
present).

The above issue has been mitigated by setting the
'default_max_wait_time' configuration option to a low threshold (0.05
seconds). This means that, even if we use a suboptimal matcher, we won't
sleep for very long.

On top of this, we now specifically disable waiting for a number of
Capybara's matches by sending the option `wait: false` into the Capybara
matcher calls. These two effects combined will mean that it should be
reasonably unlikely for the issue to recur - however we should still be
conscientious when using Capybara's matchers in the future.

It will be important to be aware, in the future, that when we do add
further dynamic content/ajax-y stuff to the frontends, that we'll
specifically need to give a generous `wait` for the matchers (for which
we can use the new `dm_custom_wait_time` config value defined in
`features/support/env.rb`

Bump Capybara library version - required for correct passing of
`options`/`wait: false` down Capybara's matcher chain.

## Before
```
218 scenarios (1 skipped, 217 passed)
2622 steps (25 skipped, 2597 passed)

Took 847 seconds (14:07)
```

## After
```
218 scenarios (1 skipped, 217 passed)
2621 steps (25 skipped, 2596 passed)

Took 375 seconds (6:15)
```

🎉 🎉 🎉  These will probably run even faster on the Jenkins box. 🎉 🎉 🎉

### DIFF
--- a/features/buyer/requirements.feature
+++ b/features/buyer/requirements.feature
@@ -137,6 +137,7 @@ Scenario: Create user research participants
    And I don't see the 'Review and publish your requirements' link
 
 
+@copy-requirements
 Scenario Outline: Copy requirements
   Given I have a live digital-outcomes-and-specialists framework
   And I have a buyer user
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -161,10 +161,10 @@ end
 When /I choose #{MAYBE_VAR} radio button(?: for the '(.*)' question)?$/ do |radio_label, question|
   if question
     within(:xpath, "//span[normalize-space(text())='#{question}']/../..") do
-      choose_radio(radio_label)
+      choose_radio(radio_label, wait: false)
     end
   else
-    choose_radio(radio_label)
+    choose_radio(radio_label, wait: false)
   end
 end
 
@@ -176,7 +176,7 @@ end
 When /I choose a random '(.*)' radio button$/ do |name|
 
   radio = all_fields(name, type: 'radio').sample
-  choose_radio(radio)
+  choose_radio(radio, wait: false)
 end
 
 When /I check all '(.*)' checkboxes$/ do |checkbox_name|
@@ -217,9 +217,9 @@ end
 # the rescue is used because if a banner cannot be found the function throws an exception and does not look for the other option
 Then /^I see a (success|warning|destructive|temporary-message) banner message containing '(.*)'$/ do |status, message|
   begin
-    banner_message = page.find(:css, ".banner-#{status}-without-action")
+    banner_message = page.find(:css, ".banner-#{status}-without-action", wait: false)
   rescue Capybara::ElementNotFound => e
-    banner_message = page.find(:css, ".banner-#{status}-with-action")
+    banner_message = page.find(:css, ".banner-#{status}-with-action", wait: false)
   end
   expect(banner_message).to have_content(message)
 end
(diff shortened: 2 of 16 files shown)


features/buyer/requirements.feature
```diff
@@ -137,6 +137,7 @@ Scenario: Create user research participants
    And I don't see the 'Review and publish your requirements' link
 
 
+@copy-requirements
```
CrystalPea: do we need this tag here?

features/support/env.rb
```diff
@@ -68,5 +69,38 @@ def dm_notify_api_key
   ENV['DM_NOTIFY_API_KEY']
 end
 
+def dm_custom_wait_time
+  5
+end
+
 Capybara.asset_host = dm_frontend_domain
 Capybara.save_path = "reports/"
+Capybara.default_max_wait_time = 0.05
```
katstevens: Out of curiosity, what was the default before?

features/supplier/view_and_edit_g_cloud_services.feature
```diff
@@ -54,6 +54,7 @@ Scenario: Supplier user can edit the features and benefits of a service
     | field                         | value                                                                                  |
     | Service features and benefits | Service features Feature 1 New Feature 2 Service benefits Benefit 1 Updated Benefit 2  |
 
+@file-upload
```
CrystalPea: do we need those tags?

features/step_definitions/common_steps.rb
```diff
@@ -405,18 +405,12 @@
...
 Then /^I (don't |)see '(.*?)'(?: or '(.*)')? text on the page/ do |negative, expected_text, alternative_expected_text|
-  has_text = page.has_content?(expected_text)
+  method = negative.empty? ? :has_content? : :has_no_content?
 
-  has_alternative_text = false
-  if alternative_expected_text
-    has_alternative_text = page.has_content?(alternative_expected_text)
-  end
+  result = page.send(method, expected_text)
+  alternative_result = alternative_expected_text ? page.send(method, alternative_expected_text) : false
```
CrystalPea: Neat logic changes here, well done. I would only check with @brenetic style-wise about the use of ternary operators.

features/support/env.rb
samuelhwilliams: 2 seconds. So not dreadful, but that did make it more than 50% of our test run time 😨

features/buyer/requirements.feature
samuelhwilliams: We don't need it (or the file-upload one before), but I don't feel like more tags are a bad thing. :)

PR #537, opened 2018-06-12
Dh 83 functional tests for direct award flow

DilwoarH:


lfdebrux:
Sorry for the nitpick, but would a better name for the feature be `tell-us-about-contract`?

Otherwise it looks good!

features/step_definitions/direct_award_steps.rb
```diff
@@ -7,3 +7,69 @@
...
+    And I click the 'End search and continue' button
+    Then I am on the '#{search_name}' page
+  }
+end
+
+When (/^I have downloaded the search results as a file of type '(.*)'$/) do |file_type|
+  steps %{
+    And I click the 'Download search results' link
+    And I am on the 'Download your search results' page
+    And I click the 'Download search results as a spreadsheet' link
```
samuelhwilliams: This step purports to let you define the filetype you want to download, but only downloads it as a spreadsheet, so when you reference the parent step you'll always need to pass in `ods`, won't you?

features/step_definitions/direct_award_steps.rb
```diff
@@ -7,3 +7,69 @@
...
+    Given I am on the 'Did you award a contract for ‘my cloud project’' page
+    And I choose the 'No' radio button
+    And I click 'Save and continue'
+    Then I am on the 'Why didn’t you award a contract?' page
+    And I choose the 'The work has been cancelled' radio button
+    And I click 'Save and continue'
+  }
+end
+
+When(/^I do not award the contract because there are no suitable services$/) do
```
samuelhwilliams: These two steps could probably be one - the only difference is the radio button selected, which could be passed into the step.

E.g `I do not award the contract because (the work has been cancelled|there were no suitable services)`.

samuelhwilliams: Also, the name of the project being hard-coded in these steps is odd. Can we use `#{search_name}`?

features/step_definitions/direct_award_steps.rb
andrewchong: You are correct. I have made the change so that it now allows filetype to be defined and downloads the appropriate file type

features/step_definitions/direct_award_steps.rb
andrewchong: I have now changed it so that the two steps are now one.

Will need some help with persisting the name of the project for reuse. Is there an example of this somewhere that I can follow?

PR #540, opened 2018-06-19
Make steps for visiting and checking page sound different

CrystalPea:
This has been done to avoid confusion.

Trello ticket: https://trello.com/c/iHSaU4d1/454-change-syntax-for-flow-starting-on-a-page-in-functional-tests-so-it-differs-more-from-syntax-checking-if-the-user-is-on-the-righ

### DIFF
--- a/features/admin/access_a_service.feature
+++ b/features/admin/access_a_service.feature
@@ -8,14 +8,14 @@ Scenario: Admin with Service Manager role can edit, remove and publish a service
   Given I am logged in as the production admin-ccs-category user
   And I am on the 'Admin' page
   When I click 'Edit suppliers and services'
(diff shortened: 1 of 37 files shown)


katstevens:
I'm not sure about removing 'the' - it reads a little funny.

Suggest we reword the step that looks for the url to `I am at the <whatever> url` (e.g. `I am at the /admin url`), then allow the h1 step to use 'the'.

katstevens:
To clarify, we have three step definitions available:

- `I am on the 'My Amazing Page' page` (checks the `<h1>` content)
- `I visit the '/my/amazing/page' page` (goes to the url, via an HTTP request)
- `I am at the '/my/amazing/page' url` (checks we're already on the url)

PR #542, opened 2018-06-20
Assert the supplier name, not the invited user name

katstevens:
When an invited supplier creates their account, they are taken to their dashboard at `/suppliers`. The `<h1>` here is the supplier name, not the user account name.

This failure wasn't picked up in pre-merge checks as we were ignoring Notify failures (which happened before this step) - lesson learned!


samuelhwilliams:
Passes locally 👍 
screen shot 2018-06-20 at 11 31 14

PR #551, opened 2018-07-01
Cast to int before sorting

samuelhwilliams:
## Summary
Need to cast the G-Cloud framework iteration to an integer before
sorting otherwise G-Cloud 9 is considered a more recent framework than
G-Cloud 10.


galund:
You'd have thought we might have a date field we could use instead but eh :)

PR #552, opened 2018-07-02
Update page title we check for when user answers VAT question

CrystalPea:
This follows change in the page title content.


risicle:
Oh I see, so the title change got reverted...

PR #555, opened 2018-07-02
Update content for save search flow

samuelhwilliams:
I think this is all that needs updating?


features/step_definitions/direct_award_steps.rb
```diff
@@ -16,7 +16,7 @@
     Then I click the '#{search_name}' link
     Then I am on the '#{search_name}' page
     And I click the 'Export your results' link
-    Then I am on the 'End your search' page
+    Then I am on the 'Before you export your results' page
```
katstevens: This is the right text, but do we need a step before this that will reduce the number of results to below 30, so that the 'End search and continue' button will be clickable?

galund:
We're covering this in https://github.com/alphagov/digitalmarketplace-functional-tests/pull/546

PR #568, opened 2018-07-26
Revert "Add temporary @skip-staging tag to tell_us_about_contract"

lfdebrux:
I added some `@skip-staging` tags to some new functional tests from @risicle a while ago, and forgot to remove them once that feature was deployed.


risicle:
Cool, but I didn't write these FTs. Think it was you & andrew actually.

PR #570, opened 2018-08-17
Scenario for incomplete supplier application

katstevens:
Trello: https://trello.com/c/Virw5sGZ/114-500-error-question-not-found

A lack of test coverage for this feature contributed to the 500 errors appearing. 

This scenario should confirm that a supplier can resume an incomplete application. Clicking the link on the dashboard will send them to the start of the flow, and any previous answers should be pre-filled.


katstevens:
Indeed - we should have added coverage for this at the time, but unfortunately functional tests got put to the bottom of the priority pile 😿

PR #572, opened 2018-08-30
search pages: add test fetching highest possible page number of an index

risicle:
This is important to check because it is handled slightly differently from the "absurdly beyond the number of search results" case in the search api and is the largest page number a user should expect to be able to fetch and get real results. A smoke test so that we can be sure our mapping config is always in concordance with the number of items we currently have.


risicle:
p.s. this was https://trello.com/c/nZWfdHgD/427-improve-last-page-fetching-functional-tests

PR #573, opened 2018-09-26
add tests of antivirus scanning

risicle:
https://trello.com/c/RVmfbGgE/384-virus-scanning-functional-tests

~~~This isn't quite finished yet - I probably need to update the readme with information about aws credentials etc., but I'm opening this PR so people can see how I've had to jumble this lot around to get this to work.~~~ More specific information about changes & rationale is present in the changeset comments, which are probably worth looking at.

The assertion it makes so far is just the existence of the email sent record corresponding to the hash of the test file. Doesn't assert the contents of that email or the destination. Is that ok? The destination is probably quite hard to assert as it gets hijacked by the "dummy address" system present in our `DMNotifyClient`.

Should it maybe also assert the presence of the result tag on the s3 object once it has found the message? It *could*...

The process seems to only take about 10 seconds (tends to find the message on its first notify poll), so maybe it *is* a candidate for running as a smoke test? Then again this is me testing it with a nearly zero load, so that could possibly degrade at certain points in the lifecycle... :man_shrugging:


lfdebrux:
Testing just that the email is sent seems perfectly correct to me.

Running as a smoke test seems a bit unnecessary to me. Enough just to ascertain that the app is running.

features/step_definitions/antivirus_steps.rb
```diff
@@ -0,0 +1,28 @@
+require 'aws-sdk-s3'
+
+Given /^I have a randomized file continaing the eicar test signature$/ do
```
katstevens: Typo `continaing`

features/antivirus/antivirus.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Antivirus scanning
+
+Scenario: Uploading an infected file to a bucket should generate a warning email
+  Given I have a randomized file continaing the eicar test signature
+  When I upload that file to the documents bucket under the key 'functional_tests/bad.pdf'
+  Then I receive a notification regarding that file within 4 minutes
```
katstevens: Is the plan to have a non-infected file upload scenario as well, where no email is sent? 

I'm not sure about having this as a smoke test, unless we can clean up the 144 random files per day being uploaded to our documents buckets for each stage?

features/step_definitions/antivirus_steps.rb
```diff
@@ -0,0 +1,28 @@
...
+  @file = Fixtures.eicar_test_signature + SecureRandom.base64(129)
+end
+
+When /^I upload #{MAYBE_VAR} to the documents bucket under the key #{MAYBE_VAR}$/ do |file_contents, destination_key|
+  s3 = Aws::S3::Resource.new(region: 'eu-west-1')  # actual region is unimportant
+  @s3_obj = s3.bucket(dm_documents_bucket_name).object(destination_key)
+  @s3_obj_response = @s3_obj.put(
```
katstevens: Does Jenkins definitely have `put` permissions for the bucket, and do we need any permissions to be able to retrieve it later - I only ask because of my trauma in this area this past week 😸

config/example.sh
```diff
@@ -31,3 +33,5 @@ export DM_PRODUCTION_ADMIN_FRAMEWORK_MANAGER_USER_EMAIL=${DM_PRODUCTION_ADMIN_FR
 export DM_PRODUCTION_ADMIN_FRAMEWORK_MANAGER_USER_PASSWORD=${DM_PRODUCTION_ADMIN_FRAMEWORK_MANAGER_USER_PASSWORD:=Password1234}
 
 export DM_PAGINATION_LIMIT=${DM_PAGINATION_LIMIT:=5}
+
+export DM_DOCUMENTS_BUCKET_NAME=${DM_DOCUMENTS_BUCKET_NAME:=digitalmarketplace-documents-${DM_ENVIRONMENT}-${DM_ENVIRONMENT}}
```
katstevens: As discussed offline, this will need to go into the Jenkins job template.

config/example.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
-export DM_API_DOMAIN=${DM_API_DOMAIN:=http://localhost:5000}
+export DM_ENVIRONMENT=${DM_ENVIRONMENT:="preview"}
```
katstevens: As discussed offline - not sure about this. I'd rather have a separate `config/preview.sh.example` file or better yet, get rid of this example file entirely and find a way to plug in the creds repo to run locally with minimal faff (something for firebreak, maybe...)

features/antivirus/antivirus.feature
lfdebrux: Agree with @katstevens on both points

risicle: I thought about this when I was going to sleep the other night actually, and I concluded "look, if it were sending spurious emails even about clean uploads, we'd damn well know about it because we'd be getting spammed with them".

I just feel a bit nervous about leaving production un-smoke-tested, because a whole *bunch* of infrastructure things need to be in place and applied correctly for the whole loop to be working. Without this check we'd never realize if our SNS subscription got accidentally removed by a terraform apply or altered, re-added and never managed to complete the SubscriptionConfirmation process. Anything like that.

As for the 144 files, the severity of that is slightly reduced by them all being under the same key, but yeah, running so frequently would feel a bit silly. This is why I think there's a need for some kind of reduced-frequency smoke test...?

One of the other motivations I had for having the smoke tests performing this is to keep the avapi's logs full. Without this constant trickle, we're unable to do the HG "no logs" alert.

katstevens: Good point about the HG alerts. I'm happy for a new `@smoulder` tag and job to exist - once per hour seems manageable?

features/step_definitions/antivirus_steps.rb
risicle: This is a good question to which I don't know the answer. @Wynndow ?

config/example.sh
risicle: :+1:

config/example.sh
risicle: Yup, maybe, looks like all of us use this file slightly differently. Guess what we could do is have a `common` file which is basically the same as this file and a `staging.sh.example` which simply defines `DM_ENVIRONMENT` then includes `common` and a `local.sh.example` which overrides a number of things (then includes `common`)... though it all feels a lot like going back to how it *was*.

README.md
```diff
@@ -119,6 +120,13 @@ This will allow the functional tests to assert that an email has arrived in the
 
 Remember not to commit the change to `config.py`!
 
+#### AWS credentials in functional tests
+
+Tests with the tag `@requires-aws-credentials` require the environment to be set up with an AWS key that has permission
```
katstevens: I like this tag a great deal 👍 👍 👍

config/common.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
-export DM_API_DOMAIN=${DM_API_DOMAIN:=http://localhost:5000}
+export DM_ENVIRONMENT=${DM_ENVIRONMENT:="development"}
```
katstevens: Does `DM_ENVIRONMENT` need to be in common? We're setting the default to `development` in `env.rb` anyway?

risicle: I put it there so that the definitions following it in the same file reusing `DM_ENVIRONMENT` don't fail to blank spaces, which can be confusing and hard to debug.

risicle:
Meh, they're supposed to be `example` files - the `preview` one should be enough to get the gist, having a `staging` one wouldn't tell them anything new.

PR #577, opened 2018-10-02
141-not-use-envrc

benvand:
Update readme information now we arent using .envrc files

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ An alternative is Docker Compose. It uses the latest tagged Docker images of the
   - Do all of this without any apps running locally to avoid issues with blocking ports.
   - Have Docker installed on your machine - see [here](https://www.docker.com/docker-mac)
   - Have Docker Compose installed on your machine. Docker for Mac comes with it so you probably won't need to do anything extra. You can check with `docker-compose -v`. If you don't have it, go [here](https://docs.docker.com/compose/install/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,7 +143,7 @@ An alternative is Docker Compose. It uses the latest tagged Docker images of the
...
   - Have Docker installed on your machine - see [here](https://www.docker.com/docker-mac)
   - Have Docker Compose installed on your machine. Docker for Mac comes with it so you probably won't need to do anything extra. You can check with `docker-compose -v`. If you don't have it, go [here](https://docs.docker.com/compose/install/)
-  - Have a `.envrc` file in the root of your credentials repo which sets your AWS profile to your sops profile. Probably something like `AWS_PROFILE=sops`. Check in `~/.aws/config` if you're not sure. This step isn't strictly necessary to get all the containers up and running, but functional tests will fail whenever trying to send an email if you don't do it.
+  - Have a valid `sops` profile in your `~/.aws/config`. This step isn't strictly necessary to get all the containers up and running, but functional tests will fail whenever trying to send an email if you don't do it. See the [set up AWS credentials](https://github.com/alphagov/digitalmarketplace-credentials/blob/master/README.md#set-up-aws-credentials-for-sops) section of the credentials repo README.md for more information.
```
katstevens: I don't know if this bit about the emails is actually right anyway - to send eg Notify emails when running functional tests locally, you need a sandbox `NOTIFY_API_KEY` in the config of the app trying to send an email (see line 100 of this README).

benvand: screen shot 2018-10-03 at 09 50 13


It's referring to this bit of the Makefile I think. `sops-wrapper` looks for a `sops` profile and the creds won't decrypt without it but it'll still succeed in setting up the container.

katstevens: Interesting! That bit never worked for me 😹

PR #581, opened 2018-10-25
Add a step to visit a supplier service page

lfdebrux:
The background for the view and edit G-Cloud services feature was
by necessity specific to which-ever G-Cloud service was live, which
caused flakiness.

This PR creates a specific step which visits the supplier service
page directly by creating the page url, which means it can be easily
parametrised.

### DIFF
--- a/features/supplier/view_and_edit_g_cloud_services.feature
+++ b/features/supplier/view_and_edit_g_cloud_services.feature
@@ -13,7 +13,9 @@ Background:
   # The following step only works by virtue of there only being a single service for this supplier - multiple services on
   # multiple frameworks will cause multiple "View services" links to be present
   And I click 'View services'
-  Then I am on the 'G-Cloud 10 services' page
+  # Two steps below so we can assert the page title for either G9 or G10
+  Then I am on the 'Your G-Cloud' page
+  And I see the page's h1 ends in 'services'
   When I click 'Test cloud support service'
   Then I am on the 'Test cloud support service' page
 


features/step_definitions/supplier_steps.rb
```diff
@@ -51,3 +51,7 @@
 Given /^that supplier has a service on the (.*) lot(?: for the (.*) role)?$/ do |lot_slug, role_type|
   @service = create_live_service(@framework['slug'], lot_slug, @supplier["id"], role_type)
 end
+
+Then 'that supplier visits their page for that service' do
+  page.visit("#{dm_frontend_domain}/suppliers/frameworks/#{@framework['slug']}/services/#{@service['id']}")
```
risicle: It might make this more reusable if you dug the framework slug out of `@service` here instead of assuming that you've also got `@framework` set.

features/supplier/view_and_edit_g_cloud_services.feature
```diff
@@ -9,13 +9,7 @@ Background:
...
   And that supplier returns a signed framework agreement for the framework
   And that supplier has a service on the cloud-support lot
-  When I visit the /suppliers page
-  # The following step only works by virtue of there only being a single service for this supplier - multiple services on
-  # multiple frameworks will cause multiple "View services" links to be present
-  And I click 'View services'
-  Then I am on the 'G-Cloud 10 services' page
-  When I click 'Test cloud support service'
-  Then I am on the 'Test cloud support service' page
+  Then that supplier visits their page for that service
```
risicle: A couple of things:

 - This does end up testing _much_ less in magically jumping straight to the target page. We usually try to limit the number of assumptions of urls we jump to and instead navigate to them "naturally".
 - `Then` steps are generally intended to be used with assertions. This step doesn't really assert anything other than getting a 200 for a request to the target page, does it?

katstevens: I'm a bit reluctant to remove the `Then I am on the 'G-Cloud 10 services' page` step entirely - I think this is the only scenario where we visit the page with the supplier's list of services. We should follow the journey that the user would take (I know this isn't consistently done in other scenarios!) wherever possible.

What if we changed the step to be `Then I am on the 'Your G-Cloud' page` combined with `And I see the page's h1 ends in 'services'`?

lfdebrux:
@katstevens thanks for your commits. Your solution is definitely has the advantage of being more similar to the user flow.

If people are happy I can squash all of these commits so the history is cleaner?

PR #582, opened 2018-11-02
Remove VAT from supplier journey

katstevens:
Trello: https://trello.com/c/34Z5Oac9/216-remove-vat-number-from-supplier-journey

Added the `@skip-staging` as a separate commit so we can revert it when we deploy to staging.


features/supplier/supplier_account.feature
```diff
@@ -6,6 +6,7 @@ Background:
   And that supplier is logged in
   And I visit the /suppliers page
 
+@skip-staging
```
lfdebrux: :+1:

PR #586, opened 2018-11-14
Initial creation of smoulder tests

risicle:
https://trello.com/c/y3rcfYeG/213-create-smoulder-tests-to-allow-more-thorough-tests-notably-antivirus-tests-to-be-run-on-production

This shouldn't be merged until jenkins is ready to start running smoulder tests.

But I'm posting this so people can get an idea of the plan for what the split should be and to object.


Makefile
```diff
@@ -3,6 +3,12 @@ DM_ENVIRONMENT ?= local
 smoke-tests: setup
 	[ -f config/${DM_ENVIRONMENT}.sh ] && . config/${DM_ENVIRONMENT}.sh ; bundle exec cucumber --strict --tags @smoke-tests --tags ~@skip --tags ~@skip-${DM_ENVIRONMENT}
 
+smoulder-tests: setup
+	[ -f config/${DM_ENVIRONMENT}.sh ] && . config/${DM_ENVIRONMENT}.sh ; bundle exec cucumber --strict --tags @smoulder-tests,@smoke-tests --tags ~@skip --tags ~@skip-${DM_ENVIRONMENT}
+
+smoulder-tests-parallel: setup
+	[ -f config/${DM_ENVIRONMENT}.sh ] && . config/${DM_ENVIRONMENT}.sh ; bundle exec parallel_cucumber features/ -n 4 -o "--strict --tags @smoulder-tests,@smoke-tests --tags ~@skip --tags ~@skip-${DM_ENVIRONMENT} ${ARGS} -p run-parallel"
```
katstevens: So is this running `smoulder` and `smoke` in parallel, or running tests within each in parallel? Do we want to include smoke tests in these two commands?

features/smoke-tests/homepage.feature
```diff
@@ -0,0 +1,14 @@
+@smoke-tests @homepage
+Feature: Check appropriate links on homepage
```
katstevens: 👍 for putting this in its own feature file

Makefile
risicle: I thought yes, in the same way the functional tests also run the smoke tests - each being a superset of the next.

PR #594, opened 2019-01-28
Remove references to production users

lfdebrux:
We have a number of pre-configured users for functional tests that are used for testing functionality for the major roles; in the tests these have been referred to as "production" users, which is misleading unless the the tests are actually running on production. This is a problem because the term production is usually associated with danger!

This PR suggests a new convention; calling them 'existing' users. Steps which refer to production users have been changed accordingly; 'a production' -> 'an existing' etc.

The default email addresses for these users (which should only ever be used for local environments) have also been changed to make it clear that they are artefacts of the functional tests.

While technically these users may not exist the first time the functional tests are run, I think this is a more accurate and less alarming way of referring to these users.


lfdebrux:
> (and possibly the visual regression tests - not sure about that one)

Ah interesting, I'm tempted to change the visual regression tests as well.

> I wonder what's the best order to merge them in?

I think the Jenkins changes will have to be merged first, or at least pushed to Jenkins, as I think the functional tests get run when changes to this repo get merged.

katstevens:
Thinking about it, we could have temporarily 2 sets of vars in the Jenkins config - that way we won't have to pause smoke/smoulder tests. I think that would work?

lfdebrux:
> Thinking about it, we could have temporarily 2 sets of vars in the Jenkins config - that way we won't have to pause smoke/smoulder tests. I think that would work?

Yes we could do that. Would we want to commit that to git?

katstevens:
I'd say yes, even though it's temporary - then we have a log of what we did if we needed to do it again in the future.

lfdebrux:
> I'd say yes, even though it's temporary - then we have a log of what we did if we needed to do it again in the future.

I've changed alphagov/digitalmarketplace-jenkins#187 to fit with this.

PR #602, opened 2019-02-07
Fix functional and smoke tests

lfdebrux:
I think this should do work?

### DIFF
--- a/features/admin/access_a_service.feature
+++ b/features/admin/access_a_service.feature
@@ -4,6 +4,7 @@ Feature: Admins viewing, editing, removing and publishing supplier services
 Background:
   Given I have a published service on a live G-Cloud framework
 
+@skip-preview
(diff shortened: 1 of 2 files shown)


features/admin/access_a_service.feature
```diff
@@ -30,6 +55,24 @@ Scenario: Admin with Service Manager role can edit, remove and publish a service
...
+  And I don't see the 'Edit' link
+  And I don't see the 'Remove service' link
+
+  Examples:
+    | role                    | link_name                                 |
+    | admin-framework-manager | View suppliers and services               |
+    | admin                   | Edit supplier accounts or view services   |
+
+
+@skip-staging @skip-production
```
katstevens: We won't need `skip-production` for this one as it's not a smoke test.

lfdebrux: Ah yes

PR #605, opened 2019-02-14
Add functionality to tests assets domain and test for favicon

benvand:
Tests:

https://trello.com/c/w6gv9OJ0/376-the-404-chain-reaction-browsers-requesting-favicons-from-our-assets-host
https://github.com/alphagov/digitalmarketplace-router/pull/47

Assets domain should now 200 for a favicon

### DIFF
--- a/features/public/assets.feature
+++ b/features/public/assets.feature
@@ -1,0 +1,8 @@
+@assets @skip-local
+Feature: Favicon.ico should return a blank, transparent gif on assets. domain
+
+Scenario: Request favicon from assets domain
+  Given I visit the assets /favicon.ico page
+  Then the response code should be 200
+  And the response header content_type should be image/gif
+  
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -20,6 +20,14 @@ Given /^I (?:re-?)?visit the (.* )?(\/.*) page$/ do |app, url|
   end
 end
 
+Given /^the response (?:(header ))?(.*) should be (.*)$/ do |header, attribute, value|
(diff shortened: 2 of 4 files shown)


features/public/assets.feature
```diff
@@ -0,0 +1,7 @@
+@assets @skip-local
+Feature: Favicon.ico should return a blank, transparent gif on assets. domain
+
+Scenario: Request favicon from assets domain
+  Given I visit the assets /favicon.ico page
+  Then the response code should be 200
+  And the response header content_type should be image/gif
```
jonodrew: Does this need a new line at the end of the file?

features/step_definitions/common_steps.rb
```diff
@@ -20,6 +20,14 @@
   end
 end
 
+Given /^the response (?:(header ))?(.*) should be (.*)$/ do |header, attribute, value|
+  if header
+    expect(@response.headers[attribute.to_sym].to_s).to eq(value)
```
jonodrew: This line seems odd to me. `.to_sym` converts a string to a symbol, which you then convert back to a string with `.to_s`. I think it should work as just `@response.headers[attribute].to_s`

benvand: So the attribute arg _appears_ to need to be `to_sym`'d to be used as an accessor on the headers hash
Found it here:
https://stackoverflow.com/a/25538021/3967444

There may well be another way to do it 🤔

jonodrew: Yup, you're right. I was reading the order of operations wrong. Head full of cotton wool today!

PR #606, opened 2019-02-14
Company details feature

katstevens:
Trello: https://trello.com/c/3ZUU9xIn/340-supplier-details-admin-page

Coverage for the new company details page (see https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/484).

Setup required for the new feature:
- Set up a scenario where a supplier has a service on a live framework, and we can search for their details via tabbed search
- Add some fake declaration data that we're going to display
- Allow summary table assert methods to check multi-line values (for the address)
- Allow summary table assert methods to skip `<ANY>` values

Updated some other scenarios to take into account that a) the supplier name is now a link b) updating supplier name redirects to the details page c) some permissions have changed.

### DIFF
--- a/features/admin/invite_supplier_contributor.feature
+++ b/features/admin/invite_supplier_contributor.feature
@@ -9,7 +9,7 @@ Scenario Outline: Correct users can invite a contributors to a supplier account
   And I click the 'Edit supplier accounts or view services' link
   And I enter 'DM Functional Test Supplier - Invite a contributor feature' in the 'supplier_name' field and click its associated 'Search' button
   And I click the summary table 'Users' link for the 'DM Functional Test Supplier - Invite a contributor feature' link
-  When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
+  When I enter 'functional-test-supplier-contributor@example.gov.uk' in the 'Email address' field
   And I click the 'Send invitation' button
   Then I see a success banner message containing 'User invited'
 
@@ -48,7 +48,7 @@ Scenario Outline: Correct users can invite a contributors to a supplier account
   And I click the 'Edit supplier accounts or view services' link
   And I enter 'DM Functional Test Supplier - Invite a contributor feature' in the 'supplier_name' field and click its associated 'Search' button
   And I click a summary table 'Users' link for 'DM Functional Test Supplier - Invite a contributor feature'
-  When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
+  When I enter 'functional-test-supplier-contributor@example.gov.uk' in the 'Email address' field
   And I click the 'Send invitation' button
   Then I see a success banner message containing 'User invited'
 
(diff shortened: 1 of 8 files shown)


features/admin/invite_supplier_contributor.feature
```diff
@@ -27,10 +27,40 @@ Scenario Outline: Prohibited user roles cannot manage supplier users
...
+
+@requires-credentials @notify @skip-preview @skip-local
+Scenario Outline: Correct users can invite a contributors to a supplier account
+  Given I am logged in as the existing <role> user
+  And I have a supplier with:
+    | name          | DM Functional Test Supplier - Invite a contributor feature |
+  And I click the 'Edit supplier accounts or view services' link
+  And I enter 'DM Functional Test Supplier - Invite a contributor feature' in the 'supplier_name' field and click its associated 'Search' button
+  And I click a summary table 'Users' link for 'DM Functional Test Supplier - Invite a contributor feature'
+  When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
```
risicle: Can we stick to the "safe" domains mentioned in https://alphagov.github.io/digitalmarketplace-manual/emails.html#live-stages ? It just makes all our stuff less notify-specific (and also emails to `simulate-delivered@notifications.service.gov.uk` don't actually show up in the "sent" log so there's a limit to how much we can test things against them (I know that doesn't make a difference in this specific case)

features/admin/invite_supplier_contributor.feature
```diff
@@ -1,14 +1,14 @@
...
+@requires-credentials @notify @skip-staging
 Scenario Outline: Correct users can invite a contributors to a supplier account
   Given I am logged in as the existing <role> user
   And I have a supplier with:
     | name          | DM Functional Test Supplier - Invite a contributor feature |
   And I click the 'Edit supplier accounts or view services' link
   And I enter 'DM Functional Test Supplier - Invite a contributor feature' in the 'supplier_name' field and click its associated 'Search' button
-  And I click a summary table 'Users' link for 'DM Functional Test Supplier - Invite a contributor feature'
+  And I click the summary table 'Users' link for the 'DM Functional Test Supplier - Invite a contributor feature' link
   When I enter 'simulate-delivered@notifications.service.gov.uk' in the 'Email address' field
```
risicle: I know this isn't something you've changed, but this too could do with being one of the 3 "special" domains...

features/step_definitions/common_steps.rb
```diff
@@ -316,6 +316,16 @@
   edit_link.click
 end
 
+When /I click the summary table '(.*)' (link|button) for the '(.*)' link$/ do |link_name, elem_type, field_to_edit|
+  case elem_type
+    when 'link'
+      edit_link = page.find(:xpath, "//tr/td/*/a[normalize-space(text()) = '#{field_to_edit}']/../../..//a[contains(normalize-space(text()), '#{link_name}')]")
+    else
+      edit_link = page.find(:xpath, "//tr/td/*/a[normalize-space(text()) = '#{field_to_edit}']/../../..//input[normalize-space(@value) = '#{link_name}']")
```
risicle: If you use `string()` instead of `text()` here it'll not get tripped up when used against something that has had e.g. a `<strong>` slipped into it.

features/admin/invite_supplier_contributor.feature
lfdebrux: Will do

features/admin/invite_supplier_contributor.feature
lfdebrux: Will do

PR #607, opened 2019-02-19
Add framework clarification question tests

risicle:
https://trello.com/b/mNNzL8hA

Made a few changes in doing these:

 - Tried to further standardize where we output informational messages, especially outputting "Notify ids" of found emails and in many places printing the actual "Notify ref" that we're going looking for. This is helpful in debugging.
 - Record complete "found" emails as `@email` rather than just the body as `@email_text`. The following steps checking for this can perfectly well pull out `email.body` themselves and this also allows us to dig into various properties of the message through `MAYBE_VAR`.
 - Add basic ability to negate checks for emails.
 - Add ability to require that opportunistically-grabbed framework also has clarification questions either open or closed. Cycling one of our instances through these settings and re-running tests can allow us to test both paths, or just whichever one we're preparing to bring to production soon.

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -98,7 +98,15 @@ def set_supplier_on_framework(framework_slug, supplier_id, status)
     frameworkInterest: { onFramework: status },
     updated_by: "functional tests",
   })
-  expect(response.code).to eq(200), _error(response, "Failed to update agreement status #{supplier_id} #{framework_slug}")
+  expect(response.code).to eq(200), _error(response, "Failed to update supplier_framework status #{supplier_id} #{framework_slug}")
+end
+
+def confirm_company_details_for_framework(framework_slug, supplier_id)
+  response = call_api(:post, "/suppliers/#{supplier_id}/frameworks/#{framework_slug}", payload: {
+    frameworkInterest: { applicationCompanyDetailsConfirmed: true },
+    updated_by: "functional tests",
+  })
+  expect(response.code).to eq(200), _error(response, "Failed to confirm supplier's company details for framework #{supplier_id} #{framework_slug}")
 end
 
 def register_interest_in_framework(framework_slug, supplier_id)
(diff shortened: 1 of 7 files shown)


features/support/api_helpers.rb
```diff
@@ -98,7 +98,15 @@ def set_supplier_on_framework(framework_slug, supplier_id, status)
     frameworkInterest: { onFramework: status },
     updated_by: "functional tests",
   })
-  expect(response.code).to eq(200), _error(response, "Failed to update agreement status #{supplier_id} #{framework_slug}")
+  expect(response.code).to eq(200), _error(response, "Failed to update supplier_framework status #{supplier_id} #{framework_slug}")
+end
+
+def confirm_company_details_for_framework(framework_slug, supplier_id)
+  response = call_api(:post, "/suppliers/#{supplier_id}/frameworks/#{framework_slug}", payload: {
```
katstevens: POST or PATCH? As discussed offline, I don't know if we'd ever do this step for a supplier with an existing `frameworkInterest`, with data to overwrite. (Can we even PATCH at the `frameworkInterest` level, like we do for the `declaration`?)

risicle: > POST or PATCH?

POO. I'm going to have to rethink this one...

It's a weird situation, because even if we do a fetch-modify-post-back it likely won't work without any tinkering due to all the extra properties the serialization annotates onto `User`.

risicle: Ah - I've looked at the view closer and think we might actually be in the clear: when we have "update" endpoints they tend to do two things - call `SomeModel.update_from_json()` methods to update properties that have straightforward correspondences, but then also special-case clauses to handle some of the more "magic" pseudo-properties. `update_supplier_framework` is _entirely_ special cases. `SupplierFramework` doesn't even have an `update_from_json()` method. Any missing keys on a `POST`ed `frameworkInterest` just get their clauses skipped.

So _that_ will work - it still doesn't save us from the fact that we set `applicationCompanyDetailsConfirmed` with not-actually-complete data... but... it's not as if we don't do that already e.g. `Given 'we accepted that suppliers application to the framework'` for the steps we consider "short cuts".

risicle:
Juuuust checking the whole suite still passes before I merge...

PR #610, opened 2019-02-22
assets tests: disable temporarily until we can fix this

risicle:
@benvand will be back imminently and can take a look at why this is failing.

Once we bring it back, might actually be nice to bring it back to the smoulder tests instead so we get this checking on production.


risicle:
Ahno, I _removed_ the missing trailing line break, that's why it's showing there.

PR #619, opened 2019-03-14
add test for inactive user trying to reset their password 

risicle:
https://trello.com/c/ARkdAn0z

Slightly misleadingly titled, because this _also_ includes a moderate refactor of the user login steps to try and make them more reusable and less repetitive.

~This doesn't work yet because the feature it's testing isn't merged.~

### DIFF
--- a/features/smoke-tests/users/login.feature
+++ b/features/smoke-tests/users/login.feature
@@ -3,92 +3,44 @@ Feature: Different types of user can log in
 
 @with-supplier-user
 Scenario: Supplier user can log in
-  Given I visit the homepage
-  And I have an existing supplier user
-  When I click 'Log in'
-  Then I am on the 'Log in to the Digital Marketplace' page
-  When I enter that user.emailAddress in the 'Email address' field
-  And I enter that user.password in the 'Password' field
-  And I click the 'Log in' button
-  Then I see the 'Log out' link
+  Given I have an existing supplier user
+  Then I can log in as that user
   When I click the 'View your account' link
   Then I see that user.emailAddress as the page header context
 
 @with-buyer-user
 Scenario: Buyer user can log in
-  Given I visit the homepage
-  And I have an existing buyer user
-  When I click 'Log in'
-  Then I am on the 'Log in to the Digital Marketplace' page
-  When I enter that user.emailAddress in the 'Email address' field
-  And I enter that user.password in the 'Password' field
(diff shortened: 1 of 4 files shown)


features/smoke-tests/users/login.feature
```diff
@@ -3,92 +3,44 @@ Feature: Different types of user can log in
...
-  Given I visit the homepage
-  And I have an existing supplier user
-  When I click 'Log in'
-  Then I am on the 'Log in to the Digital Marketplace' page
-  When I enter that user.emailAddress in the 'Email address' field
-  And I enter that user.password in the 'Password' field
-  And I click the 'Log in' button
-  Then I see the 'Log out' link
+  Given I have an existing supplier user
+  Then I can log in as that user
```
katstevens: Can we combine all these into a `Scenario Outline` with the role as a param?

features/support/api_helpers.rb
```diff
@@ -364,7 +364,15 @@ def create_user(user_role, custom_user_data = {})
...
   expect(response.code).to eq(201), response.body
+
   @user = JSON.parse(response.body)["users"]
+
+  if (custom_user_data['active'] || '').downcase == "false"
+    # this can't be specified at creation time so we need to follow up with an update to the user
+    response = call_api(:post, "/users/#{@user['id']}", payload: { users: {'active': false}, updated_by: 'functional_tests' })
+    expect(response.code).to eq(200), response.body
+    @user = JSON.parse(response.body)["users"]
+  end
```
katstevens: Could we specify an inactive user in the settings (`DM_INACTIVE_BUYER_USER_EMAIL`) so that we can avoid creating one each time, or randomly pick an inactive user somehow (we have more inactive suppliers than buyers, so maybe that would be a better bet)?

features/smoke-tests/users/login.feature
risicle: The only thing that would prevent us doing is having them separately tagged, which I assumed we wanted for a reason.

features/support/api_helpers.rb
risicle: :thinking: Well, yeah we could but really we already create a new user every time we go `And I have a buyer user` or `And I have a supplier user` currently. Our APIs don't exactly make this sort of "filtering for arbitrary properties" like active-ness particularly easy.

PR #621, opened 2019-03-15
synchronization: attempt to prevent poltergeist "jumping the gun" by monkeypatching Capybara's synchronize method

risicle:
This is something we've been trying to solve since the beginning of time and have never succeeded. Capybara/poltergeist will tend to "jump the gun" after clicking a link or submitting a form and attempt to apply the check in the following step to the current page before it has even managed to start the loading procedure of the following one.

Having spent the morning piecing together https://www.w3.org/TR/html5/browsers.html#navigating-across-documents and https://www.w3.org/TR/navigation-timing/#processing-model it occurs to me that we should be able to detect this pre-load state by noticing that the `navigationStart` timestamp was set later than the `domComplete` one, the latter retaining the setting from the old pageload. In such a case, wait a bit longer with this time not counting against `Capybara.default_max_wait_time`.

The thinking here is that in our tests, there is *no* point at which we want to be asserting something about the contents of a page while it is in this state (maybe if we had an extremely interactive frontend we would, but luckily we don't). I've run this on preview (from jenkins) 5 times without failures, so it seems pretty solid...

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -92,6 +92,10 @@ def dm_custom_wait_time
   5
 end
 
+def dm_pre_load_wait_time
+  5
+end
+
 Capybara.asset_host = dm_frontend_domain
 Capybara.save_path = "reports/"
 Capybara.default_max_wait_time = 0.05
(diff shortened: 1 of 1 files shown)


benvand:
Any chance you can put this bit of the explanation in the comments or commit message?

> detect this pre-load state by noticing that the navigationStart timestamp was set later than the domComplete one, the latter retaining the setting from the old pageload.

V useful 👍

risicle:
Updated comment to include the gist of that information.

PR #625, opened 2019-03-26
Add id to company details edit feature supplier

lfdebrux:
If this feature is run in parallel it is possible that multiple
suppliers can be created. By giving the test supplier a specific id the
api helper `get_or_create_supplier` will have an unchanging reference to
the test supplier.


lfdebrux:
This might not be the right fix. Closing while I investigate further.

PR #626, opened 2019-03-26
Fixes to company details edit feature

lfdebrux:


features/support/api_helpers.rb
```diff
@@ -387,6 +387,7 @@ def get_or_create_supplier(custom_supplier_data)
   if custom_supplier_data["name"] != nil
     response = call_api(:get, '/suppliers', params: { prefix: custom_supplier_data['name'] })
     @supplier = JSON.parse(response.body)['suppliers'][0]
+    return @supplier
```
katstevens: I'm curious - what steps did this affect?

lfdebrux: It affected the background for the company details feature; it has both name and registeredName, so if there was a supplier with the correct name they would be fetched, but then discarded if there was also a supplier with the correct registered name.

PR #641, opened 2019-05-29
create_supplier: provide a default supplier address 

risicle:
https://trello.com/c/4CtHMt07

Also expect this default supplier address in tests where it should show up.

This was caused by the changes in https://trello.com/c/fY2frqFG/ only exhibiting themselves after the staging database got cleared and `get_or_create_supplier` calls had to start creating their suppliers again.

Spot the unproficient ruby programmer.

(will probably require a clearout of the preview database)

### DIFF
--- a/features/admin/company_details_edit.feature
+++ b/features/admin/company_details_edit.feature
@@ -39,7 +39,7 @@ Scenario: Admin CCS Data Controller can edit a supplier's details
     | Company registered name      | We Have Edited These Company Details Ltd.       |
     | Company registration number  | 87654321                                        |
     | DUNS Number                  | <ANY>                                           |
-    | Address                      | 10 Downing Street London AB1 2CD United Kingdom |
+    | Address                      | 14 Duke Street Dublin H3 LY5 Ireland            |
 
   When I click the summary table 'Change' link for 'Company registration number'
   Then I am on the 'Update registered company number for ‘DM Functional Test Supplier - Edit company details feature’' page
@@ -51,7 +51,7 @@ Scenario: Admin CCS Data Controller can edit a supplier's details
     | Company registered name      | We Have Edited These Company Details Ltd.       |
     | Company registration number  | 12345678                                        |
     | DUNS Number                  | <ANY>                                           |
-    | Address                      | 10 Downing Street London AB1 2CD United Kingdom |
+    | Address                      | 14 Duke Street Dublin H3 LY5 Ireland            |
 
   When I click the summary table 'Change' link for 'DUNS Number'
   Then I see 'You need to contact cloud_digital@crowncommercial.gov.uk to change a supplier DUNS number.' text on the page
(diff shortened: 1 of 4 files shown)


features/supplier/supplier_account.feature
```diff
@@ -28,7 +28,7 @@ Scenario: Supplier user can provide and change supplier details before confirmin
   Then I am on the 'Registered company name' page
   And I enter 'Toys "ᴙ" Us' in the 'Registered company name' field and click its associated 'Save and return' button
 
-  Then I click the summary table 'Answer required' link for 'Registered company address'
+  Then I click the summary table 'Change' link for 'Registered company address'
```
lfdebrux: Why is this change necessary? 🤔

features/admin/company_details_edit.feature
```diff
@@ -80,8 +80,8 @@ Scenario: Admin CCS Data Controller can edit a supplier's details
...
   And I enter '87654321' in the 'Companies House number' field and click its associated 'Save' button
   When I click the summary table 'Change' link for 'Address'
-  And I enter '10 Downing Street' in the 'Building and street' field
-  And I enter 'London' in the 'Town or city' field
-  And I enter 'AB1 2CD' in the 'Postcode' field
-  And I enter 'United Kingdom' in the 'location-autocomplete' field and click the selected autocomplete option
+  And I enter '14 Duke Street' in the 'Building and street' field
+  And I enter 'Dublin' in the 'Town or city' field
+  And I enter 'H3 LY5' in the 'Postcode' field
+  And I enter 'Ireland' in the 'location-autocomplete' field and click the selected autocomplete option
```
lfdebrux: Was changing the address necessary to make the fix work?

features/support/api_helpers.rb
```diff
@@ -267,12 +267,33 @@ def create_supplier(custom_supplier_data = {})
     }]
   }
   supplier_data.update(custom_supplier_data)
+
+  creatable_attrs = {}
+  updatable_attrs = {}
+  supplier_data.each do |key, value|
+    if %i[id companiesHouseNumber contactInformation dunsNumber description name].include? key
```
katstevens: A one line comment here about where the restriction lies might be useful (could be overkill, but with func tests we usually need all the help we can get). `The API's validation restricts supplier creation to the following subset of attributes, any other attributes must be added in a second update`. It is the validation, right - not at model level?

features/admin/company_details_edit.feature
risicle: I didn't change the address - I put a _new_ address where there wasn't one before. I used a _different_ address so it was possible to differentiate in tests where this address was sourced from.

features/support/api_helpers.rb
risicle: Yeah it's in the json schema

features/supplier/supplier_account.feature
risicle: Because the test was previously expecting this field to be blank (having not been initialized by `create_supplier` and so showing `Answer required` instead of `Change` as the link.

lfdebrux: That's really weird, how was it passing before?

risicle: It had clearly had its supplier-level address copied across from a declaration at some point. Someone testing a script?

PR #643, opened 2019-06-06
brief response creation: don't bother outputting an informational line

risicle:
This was a last second addition to my previous PR, stupid in hindsight.

It's not worth finding out what ruby's got wrong with it.


risicle:
It would seem so.

PR #649, opened 2019-06-18
Update gems

lfdebrux:
https://trello.com/c/WzzLIqiy/853-smoke-smoulder-test-failures-on-search-results-page

We're having issues with our smoke tests that we're trying to debug; we should make sure that we're not chasing closed bugs.

I've tested these bumps locally, tests ran without issues.


risicle:
I think it's extremely unlikely to be the cause/solution and I'm not sure about throwing things upside down when we're trying to pin down a problem.

lfdebrux:
> I think it's extremely unlikely to be the cause/solution and I'm not sure about throwing things upside down when we're trying to pin down a problem.

Better safe than sorry no? If it throws things upside down we can always revert. Plus it might just magically fix things ;)

risicle:
The thing about turning things upside down is it doesn't always do it obviously. It's how you lead yourself into heisenbugs.

Also specifying bundler `2.0.1` is causing me a lot of work re-generating `gemset.nix` to get a reproducible environment because it exposes a bug in RubyGems.

benvand:
Good move :100:

lfdebrux:
> Also specifying bundler `2.0.1` is causing me a lot of work re-generating `gemset.nix` to get a reproducible environment because it exposes a bug in RubyGems.

Hmm, let's update to bundler `2.0.2`

risicle:
No, bundler 2.x in general is the issue.

benvand:
@risicle crossed in the wires there, I do think we might as well get this running with up to date packages. Any fixes we make now could be obsoleted just by updating packages

> it might just magically fix things

🙏 🙏 🙏

lfdebrux:
> No, bundler 2.x in general is the issue.

What's the bug? Is downgrading bundler a suitable workaround?

risicle:
> Any fixes we make now could be obsoleted just by updating packages

It's not about that, it's about the process of determining what the problem is. Switching things around means we be sure about any of the diagnosis from before the bump.

risicle:
It looks like I can actually do a bit of a nasty hack to get it _just about_ working with this setup - let me push a commit on top of this...

PR #650, opened 2019-06-18
853-smoke-smoulder-test-failures-on-search-results-page

benvand:
We set `Capybara.default_max_wait_time` to a low threshold (0.05) to avoid waiting for a massive `default_max_wait_time` when checking that elements _do not_ exist.
Therefore if a `find` on an element raises Capybara::ElementNotFound we only wait 0.05 seconds for it to exist before returning the fact that it does not.

This is good because it means that asserting thwat an element does not exist only takes 0.05 seconds of wait.

What's bad about the previous implementation is that it also catches 2 other errors:
https://github.com/teamcapybara/capybara/blob/2.18.0/lib/capybara/node/base.rb#L125
https://github.com/teampoltergeist/poltergeist/blob/master/lib/capybara/poltergeist/driver.rb#L376
`Capybara::Poltergeist::ObsoleteNode` and `Capybara::Poltergeist::MouseEventFailed`

We are currently experiencing failures under load when only waiting 0.05 seconds on a `Capybara::Poltergeist::ObsoleteNode`.

We need to have a different wait time for Capybara::ElementNotFound and other, legitimate, page load errors.

This pull request monkeypatches Capybara::Node::Base.synchronise with a single line addition:

`seconds = e.class == Capybara::ElementNotFound ? seconds : dm_custom_wait_time`

which sets 'seconds' to our low wait time `0.05` if we're dealing with a `Capybara::ElementNotFound` error, else sets it to our standard `dm_custom_wait_time` (5s)
d229589

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -128,12 +128,45 @@ if ENV['DM_DEBUG_SLOW_TESTS']
   end
 end
 
-# Monkeypatch Capybara's synchronize method to wait longer if it detects the page to be in a loading state.
-# Note this is done *after* the possible patching via synchronize_with_timeout_error, so that it wraps
-# *outside* that.
 module Capybara
   module Node
     class Base
+      # Monkeypatch the internals of the syncronise method to wait for `dm_custom_wait_time` seconds unless the error is
+      # Capybara::ElementNotFound in which case use the `default_max_wait_time`. We require a shorter wait time on
+      # ElementNotFound in case we are checking that an element does not exist. We don't want to wait 5 seconds for it
+      # it to appear.
+      # https://github.com/teamcapybara/capybara/blob/2.18.0/lib/capybara/node/base.rb#L77
+      def synchronize(seconds = session_options.default_max_wait_time, options = {})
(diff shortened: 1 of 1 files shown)


features/support/env.rb
```diff
@@ -98,7 +98,7 @@ def dm_pre_load_wait_time
 
 Capybara.asset_host = dm_frontend_domain
 Capybara.save_path = "reports/"
-Capybara.default_max_wait_time = 0.05
+Capybara.default_max_wait_time = 5
```
lfdebrux: It's not clear from the commit message why the timeout needs to be increased.

benvand: Crossed in the wires again I think
https://github.com/alphagov/digitalmarketplace-functional-tests/compare/af20722efb4c78bf2a9049349f8e1596faf7829c..319557b5398c648b5cd779189cbd4355f89b9c79

risicle: 5 seconds is a very long time for this, and could easily lead to extremely long test runs.

PR #651, opened 2019-06-18
add a (temporary, nastily implemented) detailed debugging log output …

risicle:
Anyone's welcome to make this less nasty for me, but it should at least let us know when the `synchronize_with_unload_wait` mechanism has been engaged for a step. Inside the `synchronize_with_unload_wait` loop we output an abbreviated version of the call stack, including only paths including `/features/` - this seems a fairly reliable way to identify code under our control.

The "proper" way of doing this would probably involve adding a custom `Formatter`, but it doesn't seem particularly straightforward to send custom output to a specific formatter.


risicle:
Oh man a bit of me was hoping this wouldn't get approved...

PR #656, opened 2019-06-19
gemset.nix: revert parallel_tests to 2.15.0 to match Gemfile.lock changes

risicle:
This works, but only just - nix systems only work if you temporarily remove the `BUNDLED WITH` lines from `Gemfile.lock`.


risicle:
Also added the `report_builder` reversion trying to keep up with the churn...

PR #657, opened 2019-06-19
Update parallel_tests from 2.15.0 to 2.18.0

lfdebrux:
parallel_tests v2.18.0 is the last version to support cucumber ~> 2.0.


risicle:
FFS

PR #662, opened 2019-06-26
Catch Capybara::ExpectationNotMet (subclass of Capybara::ElementNotFo…

benvand:
…und)

In ruby, exception comparison should be done using `e.is_a?` instead of equality syntax

Also remove test 0.05s sleep for obsoleteNode

Speeds tests back up to 5mins~


risicle:
Sigh. Rubocop still not happy.

PR #668, opened 2019-06-27
Add step for filtering search results by category

lfdebrux:
https://trello.com/c/HYk2HSCJ/896-smoke-test-failures-on-production

Our catalogue tests can be tripped up if there happens to be a search result that contains the text of a filter or category in its name.

This PR adds a more specific step for filtering by category that uses a CSS selector to limit where we look for the specific category name.

This should fix the problems we were seeing yesterday, so I've reverted the commits that added skip-production tags.

### DIFF
--- a/features/smoulder-tests/supplier/opportunities.feature
+++ b/features/smoulder-tests/supplier/opportunities.feature
@@ -50,7 +50,6 @@ Scenario: Specialist roles are selectable for Digital specialists
   When I note the result_count
   Then a filter checkbox's associated aria-live region contains that result_count
 
-@skip-production
 Scenario Outline: Specialist roles are not selectable for non-Digital specialists lots
   Given I visit the /digital-outcomes-and-specialists/opportunities page
   Then I don't see a 'Designer' checkbox
@@ -64,7 +63,6 @@ Scenario Outline: Specialist roles are not selectable for non-Digital specialist
     | Digital outcomes           |
     | User research participants |
 
-@skip-production
 Scenario Outline: User gets no results for impossible combinations of location and lot
   Given I visit the /digital-outcomes-and-specialists/opportunities page
   And I click '<lot>'
@@ -87,7 +85,6 @@ Scenario Outline: User gets no results for impossible combinations of location a
 
 
 
-@skip-production
 Scenario Outline: User can filter by status, lot, location and keyword together
   Given I visit the /digital-outcomes-and-specialists/opportunities page
   When I note the result_count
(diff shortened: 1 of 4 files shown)


lfdebrux:
I've tested these locally (running against production)

features/smoulder-tests/supplier/opportunities.feature
```diff
@@ -50,12 +50,11 @@ Scenario: Specialist roles are selectable for Digital specialists
   When I note the result_count
   Then a filter checkbox's associated aria-live region contains that result_count
 
-@skip-production
 Scenario Outline: Specialist roles are not selectable for non-Digital specialists lots
   Given I visit the /digital-outcomes-and-specialists/opportunities page
   Then I don't see a 'Designer' checkbox
   And I don't see any 'specialistRole' checkboxes
-  When I click '<lot>'
+  When I limit the search results to the '<lot>' category
```
katstevens: Let's not call it `category`, as that means something else in G-Cloud searches and could get confusing. The `'<lot>' lot`?

lfdebrux: Ah right good point

features/step_definitions/catalogue_steps.rb
```diff
@@ -64,6 +64,10 @@
   puts "Noted page_count: #{@page_count}"
 end
 
+Then(/^I limit the search results to the (.*) category$/) do |category|
```
risicle: Good stuff - could you just leave a comment here to explain _why_ we've got this special case?

features/smoulder-tests/supplier/opportunities.feature
lfdebrux: Ideally this step would be general to both G-Cloud and DOS search pages though; I called it category initially because it's what it's called in the user interface (even though it is lots in DOS). If we called it lot would that get confusing for G-Cloud when it's not a lot?

features/step_definitions/catalogue_steps.rb
lfdebrux: Done

features/smoulder-tests/supplier/opportunities.feature
benvand: Can we leave it with 'click' in the name, have to dig in to the code to see what it actually does otherwise
`When I click the '<lot>' lot/category`

lfdebrux: I've changed it to `The I click the '<lot>' category link`

PR #670, opened 2019-07-04
synchronize_with_navigation_wait: add xhr_semaphore support

risicle:
https://trello.com/c/6htyRkFX

Renamed from `synchronize_with_unload_wait`, this wrapper now also looks for a `window.GOVUK.xhr_semaphore` property in the page's DOM, which, when present and non-zero, is taken to mean that the page has an ajax request in flight and the test should delay element testing.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -167,21 +167,40 @@ module Capybara
       # Note this is done *after* the possible patching via synchronize_with_timeout_error, so that it wraps
       # *outside* that.
 
-      def synchronize_with_unload_wait(*args, &block)
+      def is_page_loading
+        browser_params = driver.evaluate_script(%q({
+          "navigationStart": window.performance.timing.navigationStart,
+          "loadEventEnd": window.performance.timing.loadEventEnd,
+          "xhr_semaphore": (window.GOVUK && window.GOVUK.xhr_semaphore) || 0
+        }))
+
+        if browser_params['navigationStart'] > browser_params['loadEventEnd']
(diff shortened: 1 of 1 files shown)


features/support/env.rb
```diff
@@ -167,21 +167,40 @@ def synchronize(seconds = session_options.default_max_wait_time, options = {})
...
+          # presumably be targeted at the upcoming page, which isn't ready.
+          reason = "navigationStart (#{browser_params['navigationStart']}) > loadEventEnd (#{browser_params['loadEventEnd']})"
+        elsif browser_params['xhr_semaphore'] > 0
+          # window.GOVUK.xhr_semaphore's non-zero value indicates that the page still has an ajax request in flight
+          reason = "xhr_semaphore = #{browser_params['xhr_semaphore']}"
+        else
+          return false
+        end
+
+        DEBUG_FILE.write "#{Time.now.utc.round(10).iso8601(6)}: #{reason} @ #{caller.to_a.select { |pth| pth.include? '/features/' }}\n"
```
katstevens: Oooh - where does `DEBUG_FILE` go then? Can we see it in Jenkins somewhere?

risicle: You _would_ be able to, except it has an unpredictable filename because it includes the pid. However if you ssh in you should see at least one `debug.<n>.out` in the same directory as the rest of the reports.

risicle:
Yeah I think it would be nice if some others could try and run it, but of course, especially in this case, there's no substitute for the real thing.

PR #673, opened 2019-07-05
Revert "synchronize_with_navigation_wait: add xhr_semaphore support"

risicle:
Reverts alphagov/digitalmarketplace-functional-tests#670

I think we should pull this out for now until I've an idea why jquery is so overzealous on thinking it has ajax loading.


risicle:
omg i hate rubocop

PR #680, opened 2019-07-19
antivirus test: upload file with a weirder key name

risicle:
https://trello.com/c/l2On3zPU

To test https://github.com/alphagov/digitalmarketplace-antivirus-api/pull/47

Though I'm going to have to think of a way of checking this only on preview to begin with...


risicle:
I did find a way, through https://github.com/alphagov/digitalmarketplace-jenkins/pull/256 - and it works.

PR #681, opened 2019-07-26
add Admin acknowledging service edits feature 

risicle:
https://trello.com/c/CnowkGc5

This was originally going to be at least two different scenarios, but I realized that the setup of the "race condition" test was the same as the contents of the first test, so decided not to waste anyone's time making it do it twice.

Can anyone think of any other scenarios I could test?

### DIFF
--- a/features/admin/acknowledge_service_edits.feature
+++ b/features/admin/acknowledge_service_edits.feature
@@ -1,0 +1,44 @@
+@admin
+Feature: Admin acknowledging service edits
+
+Background:
+  Given I have a published service on a live G-Cloud framework
+  And I ensure that all update audit events for that service are acknowledged
+
+Scenario: Admin approves service edits
+  Given I choose a random sentence
+  And user 'functional.test@example.com' sets the serviceDescription of that service to that random_sentence
+  And I am logged in as the existing admin-ccs-category user
+  And I am on the 'Admin' page
+  When I click 'Review service changes'
+  Then I am on the 'Check edits to services' page
+  And I don't see a banner message
+  And I see that service.supplierName in the 'Edited services' summary table
+  When I click the summary table 'View changes' link for that service.id
+  Then I am on that service.serviceName page
+  And I see 'functional.test@example.com made 1 edit' in the page's main
+  And I see that random_sentence in the page's ins
+
+  # Before we approve the edit we're going to test a race condition of the supplier re-editing the service
+  # before approval
+
+  When I choose a random sentence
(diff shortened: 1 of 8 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -344,7 +344,7 @@
   edit_link.click
 end
 
-When /I click a summary table '(.*)' (link|button) for '(.*)'$/ do |link_name, elem_type, field_to_edit|
+When /I click a summary table #{MAYBE_VAR} (link|button) for #{MAYBE_VAR}$/ do |link_name, elem_type, field_to_edit|
```
katstevens: Possible symptom of me not looking at the functional tests enough, but I always end up having to remind myself what extra `MAYBE_VAR` gives us and why it is good. Perhaps a quick sentence in the README pointing to the `transforms.rb` would be useful (doesn't have to be in this PR?).

features/admin/acknowledge_service_edits.feature
```diff
@@ -0,0 +1,44 @@
...
+  And I ensure that all update audit events for that service are acknowledged
+
+Scenario: Admin approves service edits
+  Given I choose a random sentence
+  And user 'functional.test@example.com' sets the serviceDescription of that service to that random_sentence
+  And I am logged in as the existing admin-ccs-category user
+  And I am on the 'Admin' page
+  When I click 'Review service changes'
+  Then I am on the 'Check edits to services' page
+  And I don't see a banner message
```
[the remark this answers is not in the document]
risicle: Either. It's just a condition we want to be true, so I've included it.

risicle:
Ok fixed now - just managed to pass 100% when run on staging.

features/support/transforms.rb
```diff
@@ -4,6 +4,12 @@
 MAYBE_VAR_RE = '((?:the )?(?:\'(?<value>.*)\')|(?:that (?<quoted>quoted )?(?<variable>\w+)(?<attributes>(?:\.[\w-]+)*)))'
 MAYBE_VAR = MAYBE_VAR_RE.gsub(/<[a-z]+>/, ':')
 
+# the following transform, used in conjunction with our habit of putting quotation marks
+# *outside* our steps' capture groups, is unfortunately overzealous - causing any captured
+# parameter to be possibly transformed if e.g. beginning with "that ".
+# TODO we should decide what to do about this, possibly by making _all_ our outer-quoted
+# capture groups use MAYBE_VAR somehow
```
katstevens: 👍

PR #685, opened 2019-08-14
Fix a failing functional test that looked for the count of services

jonodrew:
Laurence and I did some digging and tweaked the 'User is able to click on several random filters' feature to look for invisible ids. This seems to have fixed a previously flaky test

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -507,7 +507,7 @@ Then(/^a filter checkbox's associated aria-live region contains #{MAYBE_VAR}$/) 
       page.all(
         :xpath,
         "//div[contains(@class, 'govuk-option-select')]//input[@type='checkbox']").sample["aria-controls"],
-        visible: :true
+        visible: :all
     ).text(:all)
   ).to include(value.to_s)
 end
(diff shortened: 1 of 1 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -506,8 +506,9 @@
     page.find_by_id(
       page.all(
         :xpath,
-        "//div[contains(@class, 'govuk-option-select')]//input[@type='checkbox']").sample["aria-controls"]
-    ).text
+        "//div[contains(@class, 'govuk-option-select')]//input[@type='checkbox']").sample["aria-controls"],
+        visible: :true
```
lfdebrux: `visible: true` is the default, I think we wanted this `visible: :all` is what we wanted for Chrome

```suggestion
        visible: :all
```

lfdebrux:
I've tested this locally and it works

features/step_definitions/common_steps.rb
```diff
@@ -506,8 +506,9 @@
     page.find_by_id(
       page.all(
```
risicle: Where's the closing bracket corresponding to this?

jonodrew:
```
page.find_by_id(
  page.all(
    :xpath,
    "//div[contains(@class, 'govuk-option-select')]//input[@type='checkbox']"
  ).sample["aria-controls"], visible: :all
).text(:all)
```
tl;dr in the middle of line 509

risicle: Ahhhhhhhhhhh

benvand:
I'm happy to take this out while @jonodrew is away.

LGTM

PR #686, opened 2019-08-20
Change service edits feature test to acknowledge all service edits

lfdebrux:
We had an issue where functional tests were failing because the number
of unacknowledge service edits on the admin page was larger than 100, so
the most recent were being pushed on to a second page and not visible.

This PR changes the service edit feature's background to make sure
that all unacknowledged service edits are acknowledged, so we can be
sure the page is clear before beginning the tests.

### DIFF
--- a/features/step_definitions/supplier_steps.rb
+++ b/features/step_definitions/supplier_steps.rb
@@ -68,6 +68,10 @@ Given /^that supplier has a service on the (.*) lot(?: for the (.*) role)?$/ do 
   puts "service id: #{@service['id']}"
 end
 
+Given 'I ensure that all update audit events are acknowledged' do
+  acknowledge_all_service_updates
+end
+
 Given 'I ensure that all update audit events for that service are acknowledged' do
   acknowledge_all_service_updates(@service['id'])
 end
--- a/features/admin/acknowledge_service_edits.feature
+++ b/features/admin/acknowledge_service_edits.feature
@@ -3,7 +3,7 @@ Feature: Admin acknowledging service edits
 
 Background:
   Given I have a published service on a live G-Cloud framework
-  And I ensure that all update audit events for that service are acknowledged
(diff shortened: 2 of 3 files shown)


features/step_definitions/supplier_steps.rb
```diff
@@ -68,6 +68,10 @@
   puts "service id: #{@service['id']}"
 end
 
+Given 'I ensure that all update audit events are acknowledged' do
+  acknowledge_all_service_updates
+end
+
 Given 'I ensure that all update audit events for that service are acknowledged' do
```
jonodrew: This step is only used in [one other place](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/76ce67b17f711f2fb904c35040b55c212161485b/features/supplier/view_and_edit_g_cloud_services.feature#L12) - is it worth changing that step to your new one too, and getting rid of this one altogether?

lfdebrux: I can't think of any reason not to... do you have any thoughts @risicle?

risicle: That would abandon any attempt at making these tests isolatable, which I did quite intentionally. It's a pity that they will eventually fall off the bottom of the 100 edits. What about making the other tests which make service edits call `I ensure that all update audit events for that service are acknowledged` when they're done?

lfdebrux: @risicle does that comment extend to the whole PR?

risicle: More or less, yes.

lfdebrux:
> That would abandon any attempt at making these tests isolatable, which I did quite intentionally. It's a pity that they will eventually fall off the bottom of the 100 edits. What about making the other tests which make service edits call I ensure that all update audit events for that service are acknowledged when they're done?

I'm going to create an alternate PR with this suggestion

PR #688, opened 2019-08-22
Allow using briefs where the user is not active

lfdebrux:
We have been running out of closed briefs with responses, partly because we limit ourselves to briefs where the user is active; this restriction is not necessary however. This commit adds a method `ensure_user_can_log_in` that activates deactivated users, and we use this in the "log in as user with brief with brief responses" step.

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -24,7 +24,8 @@ Given /^I am logged in as the buyer of a closed brief with responses$/ do
   puts "brief id: #{@brief['id']}"
   @lot_slug = @brief['lotSlug']
   @framework_slug = @brief['frameworkSlug']
-  @buyer_user = (@brief['users'].select { |u| u["active"] && !u["locked"] })[0]
+  @buyer_user = @brief['users'].sample
+  ensure_user_can_log_in @buyer_user
   puts "user id: #{@buyer_user['id']}"
   @buyer_user.update('password' => ENV["DM_BUYER_USER_PASSWORD"])
   steps "Given that buyer is logged in"
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -84,6 +84,15 @@ def ensure_user_exists(user_details)
   get_user_by_email(user_details['emailAddress'])
 end
 
+def ensure_user_can_log_in(user)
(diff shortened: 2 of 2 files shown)


features/step_definitions/brief_response_steps.rb
```diff
@@ -24,7 +24,8 @@
   puts "brief id: #{@brief['id']}"
   @lot_slug = @brief['lotSlug']
   @framework_slug = @brief['frameworkSlug']
-  @buyer_user = (@brief['users'].select { |u| u["active"] && !u["locked"] })[0]
+  @buyer_user = @brief['users'].sample
+  ensure_user_can_log_in @buyer_user
```
katstevens: Would we have to deactivate the user again afterwards - some of our scenarios test that inactive users can't do things?

lfdebrux: I didn't think it would be an issue. The only test I could find where we do this we create an inactive user; I think we don't even have a way (in our functional tests) of getting an existing inactive user.

risicle:
:thinking: 

The problem we really have here is that we're not able to do proper queries to find a usable case - we're restricted to the API. Instead of being able to ask "Give me a brief that's in state x and whose user is active" we do "Pick the first brief you find in state x and check if the user's active and oh if it isn't then - whoops - fail". We could do something a little like a proper query if we made the requests "backtrack" and try the next brief if this one's not appropriate. i.e. put it in a loop.

It would be nice if we could make the routine at least explore these avenues a bit further before it resorts to mutating the database...

lfdebrux:
@risicle I did consider creating a loop, however that would probably result in the functional test run time exploding, and not solve the issue as we have a lot of inactive users. I don't see a way around that without mutating the database.

risicle:
Could it not limit itself to e.g. 5 attempts? If we really had _that_ many inactive users, our tests would fail almost every time, because currently they give up on the first try. But they don't.

risicle:
Also, I'm suggesting mutating the database as a last resort.

lfdebrux:
> Could it not limit itself to e.g. 5 attempts? If we really had _that_ many inactive users, our tests would fail almost every time, because currently they give up on the first try. But they don't.

You're right, it's not as bad as I feared

lfdebrux:
Closing in favour of #689

PR #689, opened 2019-08-23
Change step to find briefs to use enumerator

lfdebrux:
Alternative to https://github.com/alphagov/digitalmarketplace-functional-tests/pull/688

We have been tests fail due to closed briefs with responses not having active users; this commit changes the step to try a number of brief responses until an appropriate one is found.

We currently only have 3 closed briefs with submitted responses and no active users, so this shouldn't cause any slowdown 🤞

### DIFF
--- a/features/step_definitions/brief_response_steps.rb
+++ b/features/step_definitions/brief_response_steps.rb
@@ -19,12 +19,15 @@ Given /^I am logged in as the buyer of a (closed|live) brief$/ do |status|
 end
 
 Given /^I am logged in as the buyer of a closed brief with responses$/ do
-  submitted_brief_response = get_brief_responses('digital-outcomes-and-specialists-3', 'submitted', 'closed').sample
-  @brief = get_brief(submitted_brief_response['brief']['id'])
-  puts "brief id: #{@brief['id']}"
+  submitted_brief_responses = iter_brief_responses('digital-outcomes-and-specialists-3', 'submitted', 'closed')
+  submitted_brief_responses.each do |brief_response|
+    @brief = get_brief(brief_response['brief']['id'])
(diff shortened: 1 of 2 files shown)


features/step_definitions/brief_response_steps.rb
```diff
@@ -19,12 +19,15 @@
 end
 
 Given /^I am logged in as the buyer of a closed brief with responses$/ do
-  submitted_brief_response = get_brief_responses('digital-outcomes-and-specialists-3', 'submitted', 'closed').sample
-  @brief = get_brief(submitted_brief_response['brief']['id'])
-  puts "brief id: #{@brief['id']}"
+  submitted_brief_responses = iter_brief_responses('digital-outcomes-and-specialists-3', 'submitted', 'closed')
```
katstevens: Also, do they have to be DOS3 brief responses? Could we look at DOS/DOS2 to bump up the numbers?

lfdebrux: I don't see why not 🤔

lfdebrux: We actually have reasonable numbers of closed briefs though, about 150. I don't really understand why it's failing consistently.

lfdebrux: I think what is happening is that because `get_brief_responses` gets a page at a time and then filters, we've been getting into a situation where it just happens that there is one page which has only one brief response to a closed brief, and that brief just happens to have no active users. It's quite an unlikely circumstance, but because of the pagination we reach a point where we get stuck on it.

That's my theory anyway.

risicle: Yeah that does kinda make sense

PR #692, opened 2019-09-10
live service editing: tidy up unacknowledged service update audit events after scenarios

risicle:
https://trello.com/c/ObiHdZk2 - this is getting annoying.

This fix prevents the number of unacknowledged service updates accumulating to over ~100 and overflowing the admin's single page interface to them, in turn breaking its tests.

### DIFF
--- a/features/admin/access_a_service.feature
+++ b/features/admin/access_a_service.feature
@@ -5,7 +5,8 @@ Background:
   Given I have a published service on a live G-Cloud framework
 
 Scenario: Admin with Service Manager role can edit, remove and publish a service
-  Given I am logged in as the existing admin-ccs-category user
+  Given I ensure that all update audit events for that service are acknowledged
+  And I am logged in as the existing admin-ccs-category user
   And I am on the 'Admin' page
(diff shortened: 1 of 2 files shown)


katstevens:
I think there's another service edit scenario in the `admin.access_a_service.feature`

risicle:
Well caught. Do edits by admin users get included in the service edits list?

risicle:
I've just checked - they don't. But that means I can add a check to the tests that they don't, which is what I've done.

features/admin/access_a_service.feature
```diff
@@ -17,6 +18,8 @@ Scenario: Admin with Service Manager role can edit, remove and publish a service
   Then I am on the 'About your service' page
   When I enter 'Plant-based cloud hosting' in the 'serviceName' field and click its associated 'Save and return to summary' button
   Then I am on the 'Plant-based cloud hosting' page
+  # edits by admins shouldn't result in update audit events that need acknowledgement
+  And that service has no unacknowledged update audit events
```
katstevens: 👍

PR #693, opened 2019-09-10
Change summary table steps to use Capybara Finders instead of xpath

lfdebrux:
As part of migrating to our apps to use GOV.UK Frontend buttons in forms are going to change from input elements to button elements, which breaks our tests because they currently search for input elements but not button elements.

Capybara's `find_button` method takes care of this for us, as it looks for `button` elements and `input` elements. Capybara Finders are also easier to read and less brittle; we should use them instead of xpath expressions where possible.


risicle:
(I don't understand why it's allowing you to merge this directly though it's not actually up-to-date. Repo has same branch protections as others as far as I can see.

PR #699, opened 2019-10-14
Fix issue with briefs

lfdebrux:
It seems that we have been unable to run our functional tests recently because they have been looking for DOS3 briefs, but all live briefs on preview are now DOS4.


lfdebrux:
I haven't updated the step that finds closed briefs to use DOS4 yet, because there are only 6 closed DOS4 briefs on preview, vs 325 closed DOS3 briefs.

I'll add a TODO however, for future reference.

PR #700, opened 2019-10-14
Find briefs without regard to framework

lfdebrux:
I think that specifying which framework to search when finding briefs is more trouble than its worth; if we are looking for live briefs then we should only ever find briefs for the current live framework(s), and if we're looking for closed briefs if we use the "human" search order then we'll always start looking at the most recent briefs in the most recent framework anyway.

This should make the period when we change DOS frameworks much less painful, and as a side effect might mean fewer failures due to lack of briefs in the future.

As a side note, I've split these changes into small but complete parts, so if we need to in future it should be straightforward to revert them.


lfdebrux:
I'm testing these changes running locally against preview.

lfdebrux:
Tests passed

PR #707, opened 2019-11-20
Add requirements scenario for DOS requirement preview

lfdebrux:
We are testing out a new feature on preview behind a feature flag (see alphagov/digitalmarketplace-briefs-frontend#243).

This PR creates a copy of the requirements scenario tests that account for the change in link text, so we can continue running the functional tests on preview (and locally). The existing tests run on staging and production as before.

At some point we should add functional tests for the new feature, but that can wait, whereas we should have working tests now.

### DIFF
--- a/features/buyer/requirements-preview.feature
+++ b/features/buyer/requirements-preview.feature
@@ -1,10 +1,9 @@
-@buyer @requirements
+@buyer @requirements @skip-staging @skip-production
 Feature: Create and publish a requirement
   In order to find individuals and teams that can provide the needed services
   As a buyer within government
   I want to be able to create and publish a requirement
 
-@skip-local @skip-preview
 Scenario: Create individual specialist requirement
   Given I am logged in as a buyer user
     And I have created an individual specialist requirement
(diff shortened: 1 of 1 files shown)


katstevens:
I have mixed feelings about duplicating the file for the preview env. One the one hand it's easy to remove again if/when we fully release (or remove!) the feature. On the other, it includes all the scenarios, not just the four failing ones, so we have less visibility over which scenarios 'work' on preview and which don't. Hmm!

lfdebrux:
> I have mixed feelings about duplicating the file for the preview env. One the one hand it's easy to remove again if/when we fully release (or remove!) the feature. On the other, it includes all the scenarios, not just the four failing ones, so we have less visibility over which scenarios 'work' on preview and which don't. Hmm!

We could do a hybrid where we copy the failing scenarios to a new file?

lfdebrux:
> > I have mixed feelings about duplicating the file for the preview env. One the one hand it's easy to remove again if/when we fully release (or remove!) the feature. On the other, it includes all the scenarios, not just the four failing ones, so we have less visibility over which scenarios 'work' on preview and which don't. Hmm!
> 
> We could do a hybrid where we copy the failing scenarios to a new file?

See https://github.com/alphagov/digitalmarketplace-functional-tests/compare/ldeb-fix-publish-requirements-2

katstevens:
Yep, that looks ideal - thanks @lfdebrux

PR #708, opened 2019-11-26
Update validation message step to support GOV.UK Design System

lfdebrux:
As part of migrating our frontends to use GOV.UK Design System styles
and components we are replacing instances of our validation components
with the error message component.

We want to be able to have the test working on either version, so this
commit updates the step that looks for validation messages to also look
for error messages.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -258,7 +258,7 @@ Then /^I don't see a banner message$/ do
 end
 
 Then /^I see a validation message containing '(.*)'$/ do |message|
-  validation_message = page.find(:css, ".validation-message")
+  validation_message = page.find(:css, ".validation-message, .govuk-error-message")
   expect(validation_message).to have_content(message)
 end
 


features/step_definitions/common_steps.rb
```diff
@@ -258,7 +258,7 @@
 end
 
 Then /^I see a validation message containing '(.*)'$/ do |message|
-  validation_message = page.find(:css, ".validation-message")
+  validation_message = page.find(:css, ".validation-message, .govuk-error-message")
```
katstevens: What happens if we end up finding both? (I doubt it will happen in practice, just curious)

lfdebrux: I'm not really sure... my gut feeling says it should work, because there could also be instances with multiple validation messages; but then I don't know if we test any of those.

I'll do some digging.

lfdebrux: `page.find` returns the first found, so if there are two and the first one found has the correct text, the test will pass.

lfdebrux: It's not documented, but `page.find` will raise an error if it finds more than one match :D

katstevens: OK! I'd rather it fail noisily as above than giving a false positive with the wrong message, if you see what I mean.

PR #711, opened 2019-12-03
admin: change expected text 'Edit supplier declarations' -> 'View and edit supplier declarations'

risicle:
https://trello.com/c/ApyeZYne


katstevens:
Will this need `@skip-staging` etc while we deploy?

risicle:
Was planning on doing an off-branch FT run with this when the feature hits preview, but yes I suppose it will need to be merged if we're not going to have the _regular_ tests failing and will need a tag to cope with that.

PR #713, opened 2019-12-06
Update framework application helpers with new submit buttons

katstevens:
Two scenarios in our functional tests have been failing for a while, but we didn't notice because those scenarios require the framework to be in an `open` state.

The failures are due to the changes in the submit buttons (from `input` to `button`). It's not trivial to fix, because:

- our supplier application scenario sometimes expects the user to continue through each section sequentially, and sometimes expects the user to jump back to the summary page before starting the next section. The `find_and_click_submit_button` method returns `true` or `false` depending on where we're expected to go. The trouble is, we don't know where we're expected to go before we start looking for the buttons 😿  
- the wording on the buttons is also different for declaration pages and for service pages, and imply different destination pages 

I'm confident however that these functions aren't used outside the application journey.

The changes below are full of ugly xpath statements but they do at least pass. Suggestions for improvement welcome.

### DIFF
--- a/features/step_definitions/framework_application_steps.rb
+++ b/features/step_definitions/framework_application_steps.rb
@@ -25,7 +25,7 @@ Then(/^I follow the first 'Edit' link and answer all questions on that page and 
       answer = fill_form
     end
     merge_fields_and_print_answers(answer)
-    find_elements_by_xpath("//button[contains(normalize-space(text()), 'Save and continue')] | //button[contains(normalize-space(text()), 'Save and return to declaration overview')]")[0].click
+    first(:button, text: /(Save and continue)|(Save and return to declaration overview)/).click
   end
 end
 
@@ -39,14 +39,14 @@ Then(/^I submit a service for each lot$/) do
     rescue Capybara::ElementNotFound => e
       answer_all_dos_lot_questions "Edit"
       answer_all_service_questions "Add"
-      find_elements_by_xpath("//button[contains(normalize-space(text()), 'Mark as complete')]")[0].click
+      first(:button, "Mark as complete").click
     else
       answer = fill_form
       merge_fields_and_print_answers(answer)
       click_on 'Save and continue', wait: false
       answer_all_dos_lot_questions "Edit"
       answer_all_service_questions "Answer question"
-      find_elements_by_xpath("//button[contains(normalize-space(text()), 'Mark as complete')]")[0].click
+      first(:button, "Mark as complete").click
       click_on "Back to application", wait: false
     end
 
--- a/features/step_definitions/framework_application_steps.rb
+++ b/features/step_definitions/framework_application_steps.rb
@@ -25,7 +25,7 @@ Then(/^I follow the first 'Edit' link and answer all questions on that page and 
       answer = fill_form
     end
     merge_fields_and_print_answers(answer)
-    find_and_click_submit_button
+    find_elements_by_xpath("//button[contains(normalize-space(text()), 'Save and continue')] | //button[contains(normalize-space(text()), 'Save and return to declaration overview')]")[0].click
   end
 end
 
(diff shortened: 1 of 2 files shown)


features/step_definitions/framework_application_steps.rb
```diff
@@ -39,14 +39,14 @@
     rescue Capybara::ElementNotFound => e
       answer_all_dos_lot_questions "Edit"
       answer_all_service_questions "Add"
-      find_elements_by_xpath("//input[@value='Mark as complete']")[0].click
+      find_elements_by_xpath("//button[contains(normalize-space(text()), 'Mark as complete')]")[0].click
```
lfdebrux: I think this can be simplified using the `click_button` method

```suggestion
      click_button("Mark as complete")
```

features/support/form_helpers.rb
```diff
@@ -361,18 +361,23 @@ def open_categories(categories_headings)
...
   def find_and_click_submit_button
-    submit_button = find_elements_by_xpath("//input[@class='button-save']")[0].value
-    if submit_button == 'Save and continue'
-      click_on 'Save and continue', wait: false
-      false
-    else
-      click_on submit_button, wait: false
+    # Get the submit buttons on the page, and categorise by whether we expect to redirect to the summary or not
+    save_and_return_button = find_elements_by_xpath("//button[contains(normalize-space(text()), 'Save and return')] | //button[contains(normalize-space(text()), 'Save and return to service summary')]")[0]
+    save_and_continue_button = find_elements_by_xpath("//button[contains(normalize-space(text()), 'Save and continue')]")[0]
```
lfdebrux: I think this can be simplified using the `find_button` method

```suggestion
    save_and_continue_button = find_button("Save and continue")
```

features/step_definitions/framework_application_steps.rb
katstevens: I tried doing something like that before and ended up getting redirected to the GOV.UK home page... does this pass for you when you run locally?

lfdebrux: lolwat

features/step_definitions/framework_application_steps.rb
```diff
@@ -25,7 +25,7 @@
       answer = fill_form
     end
     merge_fields_and_print_answers(answer)
-    find_and_click_submit_button
+    find_elements_by_xpath("//button[contains(normalize-space(text()), 'Save and continue')] | //button[contains(normalize-space(text()), 'Save and return to declaration overview')]")[0].click
```
lfdebrux:
```suggestion
    click_button("(Save and continue)|(Save and return to declaration overview)")
```

features/step_definitions/framework_application_steps.rb
lfdebrux: Yes it worked for me locally

features/step_definitions/framework_application_steps.rb
lfdebrux:
```suggestion
    first(:button, text: /(Save and continue)|(Save and return to declaration overview)/).click
```

features/support/form_helpers.rb
katstevens: This fails if it can't find the button, right? This helper function needs to look for both types of button in order to figure out whether it's the end of the section or not, so including this line fails on the company registration details for me.

features/step_definitions/framework_application_steps.rb
katstevens: This one works just fine thought 😸

features/support/form_helpers.rb
lfdebrux: Ah yes, good point.

This will do it `first(:button, "Save and continue", minimum: 0)`, although I'm not sure whether it is more readable though.

katstevens: We use `find_elements_by_xpath` all over the place, mostly because it has the `synchronize_with_unload_wait` fix in it. I'm happy to keep it?

PR #723, opened 2020-01-24
BUGFIX-skip-supplier-creation-for-now

benvand:
Skip buyer creation tests on preview while a workaround for the duns flow is found


lfdebrux:
This is because the functional tests need a valid DUNS number now? 🤔

PR #724, opened 2020-01-24
212-supplier-details-lookup-page

benvand:
* Update create supplier test
  * new content
  * test duns numbers


lfdebrux:
Do we need to add `@skip-staging @skip-production` tags?

benvand:
@lfdebrux Should skip this on staging and prod as the fw isn't 'coming' there yet. And hopefully this will be out by the time that happens

PR #727, opened 2020-01-29
Update tests for changes to header and footer in apps

lfdebrux:
We've recently merged [changes that remov govuk_template from the buyer frontend and replaced the header and footer with new components](https://github.com/alphagov/digitalmarketplace-govuk-frontend#55).

This has broken a few tests, as the new header uses a link instead of a form for the 'Log out' button, and [the footer refers to 'G-Cloud supplier A to Z' instead of 'A-Z'](https://trello.com/c/i4eJBbNC/274-replace-content-supplier-a-z-with-supplier-a-to-z).

### DIFF
--- a/features/public/public_opportunity.feature
+++ b/features/public/public_opportunity.feature
@@ -6,7 +6,7 @@ Scenario: Public views the publish requirements. Details presented matches what 
   And I have a buyer user
   And that buyer is logged in
   And I have a live digital-specialists brief
-  And I click the 'Log out' button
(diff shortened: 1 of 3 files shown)


lfdebrux:
I've added a commit to move `@skip-preview` tag to old test, as I'm about ready to merge alphagov/digitalmarketplace-buyer-frontend#975

lfdebrux:
Okay, I've tested this locally (against preview), going to merge

PR #728, opened 2020-01-30
276-extra-func-tests-for-supplier-sign-up-w-duns

benvand:
* Add more tests for supplier registration flow with duns number check
  * Create account
  * Duns in use
  * Duns not found

https://trello.com/c/5EjhJ351/276-extra-func-tests-for-supplier-sign-up-w-duns

Run and working against preview

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -517,3 +517,8 @@ end
 Then /^I see that the page has not been reloaded/ do
   expect(page).to have_selector("body.not-reloaded")
 end
+
+When /^I see the '(.*)' field prefilled with #{MAYBE_VAR}?$/ do |field_name, value|
+  field_element = page.find_field field_name
+  expect(field_element.value).to include(normalize_whitespace(value))
+end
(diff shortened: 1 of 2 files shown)


katstevens:
I'm having trouble testing this with local `dmrunner` as the external D&B request from the Docker image is blocked on my machine. We may have to figure out a way of stubbing out the endpoint when running locally.

Would we need something like this to run the fallback scenario anyway? We are keeping the old scenario in case D&B is unavailable - how will that test pass the moment?

benvand:
Yes we're keeping the old flow in case D&B is unavailable

lfdebrux:
> I'm having trouble testing this with local `dmrunner` as the external D&B request from the Docker image is blocked on my machine. We may have to figure out a way of stubbing out the endpoint when running locally.
> 
> Would we need something like this to run the fallback scenario anyway? We are keeping the old scenario in case D&B is unavailable - how will that test pass the moment?

I can't reproduce on my machine, connection to Dun and Bradstreet from supplier frontend seems to work for me.

lfdebrux:
Running these against dmrunner works for me.

PR #739, opened 2020-02-20
supplier creation: adapt to summary tables being converted to summary lists

risicle:
For the breakage introduced by https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1186

The approach I took here is to make all of the necessary steps targeting "summary tables" able to target "summary lists" too. Managed to avoid having to rely on any class names, even though the equivalent of a `<tr>` is now a `<div>`.

~TODO before merge: appropriate `@skip-` tags.~

### DIFF
--- a/features/support/helpers.rb
+++ b/features/support/helpers.rb
@@ -223,7 +223,7 @@ def escape_xpath(string)
 end
 
 def get_table_rows_by_caption(caption)
-  result_table_xpath = "//caption[@class='visually-hidden'][normalize-space(text())=\"#{caption}\"]/parent::table"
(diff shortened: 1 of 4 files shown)


features/support/helpers.rb
```diff
@@ -228,3 +228,11 @@ def get_table_rows_by_caption(caption)
   result_table_rows_xpath = result_table_xpath + rows_xpath
   page.all(:xpath, result_table_rows_xpath)
 end
+
+def get_summary_list_rows_by_preceding_heading(heading)
+  # do we even h5?
```
katstevens: I have grepped our frontend apps and cannot see any `h5`s.

risicle: :tada:

PR #748, opened 2020-03-05
Wait for document.readyState on search-related smoke tests on Preview

katstevens:
https://trello.com/c/xD6lS5Io/376-stop-smoke-tests-going-off-for-obsoletenode-error

**Observation**: search result pages for G-Cloud and DOS seem more susceptible to ObsoleteNode errors than other pages (looking at recent smoke, smoulder and functional test failures). 

**Hypothesis**: explicitly checking whether a search result page has finished loading before attempting to click will reduce the likelihood of Capybara being unable to find a link element. 

What I've tried:
- Insert a new step 'wait for the page to load' on scenarios where we're clicking directly on search results (i.e. not after using filters/live search - we already have a separate async handler for waiting for Ajax requests to finish)
- The page checks for the `document.readyState` property to be 'complete' (see https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState). Even though we're not waiting for Ajax requests, these pages are slow to load (see analytics info on the ticket). 
- The step will wait until the maximum `dm_custom_wait_time` (currently 5s) before giving up
- Only applied to the two Preview smoke test scenarios at present - we can experiment by merging several apps at once (a known trigger for ObsoleteNode errors) and see we have fewer failures.

### DIFF
--- a/features/smoke-tests/buyer/catalogue.feature
+++ b/features/smoke-tests/buyer/catalogue.feature
@@ -9,7 +9,6 @@ Scenario: User can click through to g-cloud page
   When I click 'Start a new search'
   Then I am on the 'Choose a category' page
 
-@skip-preview
 Scenario: User is able to navigate to service detail page via selecting the service from the search results
   Given I visit the /g-cloud/search page
   Then I am on the 'Search results' page
--- a/features/smoke-tests/buyer/catalogue.feature
+++ b/features/smoke-tests/buyer/catalogue.feature
@@ -9,9 +9,19 @@ Scenario: User can click through to g-cloud page
   When I click 'Start a new search'
   Then I am on the 'Choose a category' page
 
+@skip-preview
 Scenario: User is able to navigate to service detail page via selecting the service from the search results
   Given I visit the /g-cloud/search page
   Then I am on the 'Search results' page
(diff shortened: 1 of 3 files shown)


risicle:
What if ... you left both variants enabled for preview while we're testing it... so that we can see when one fails and one passes on the same run?

katstevens:
@risicle don't see why not.

benvand:
I think this will work.
I'm torn thinking this should be implemented on all page loads:
https://github.com/alphagov/digitalmarketplace-functional-tests/blob/9b644d774960f6261394bd481aef66cf9120573d/features/support/env.rb#L172

risicle:
If I had found that checking `document.readyState` worked well enough, do you think I would have gone to all the bother digging in to the specifications for obscure APIs like `window.performance.timing` for that line?

There's only one way to find out though...

katstevens:
> If I had found that checking `document.readyState` worked well enough, do you think I would have gone to all the bother digging in to the specifications for obscure APIs like `window.performance.timing` for that line?

My theory is that it's two different problems - the existing `synchronize` code should ('should') mitigate the async issues we had around live search. I hope that this change will help pages that aren't 'async' as such, merely slow-loading.

PR #750, opened 2020-03-12
g-cloud search tests: incorporate testing of multi-valued filter selections being preserved across pages

risicle:
https://trello.com/c/KaL0fCtR

Use security clearance filter as values are quite stable across frameworks and selecting only two of them should still give us a reliably large number of results.


risicle:
The former - just wait till it's out. Maybe run it off-branch to verify the fix on staging at release time.

risicle:
That's a point actually I haven't run it against staging/prod to check if fails with the fix unapplied.

risicle:
Now I have and it does.

PR #755, opened 2020-05-11
Update steps for changes to buyer frontend content

lfdebrux:
https://trello.com/c/v6elozgK/25-2-high-level-content-review-of-pages-in-search-for-g-cloud-services-journey

Fixes tests to pass against alphagov/digitalmarketplace-buyer-frontend#1030

To make releasing easier I've added a second commit that can be reverted once alphagov/digitalmarketplace-buyer-frontend#1030 has been released.

We want the tests to pass on both preview and staging with slightly different content for the checkbox 'I understand that I cannot edit my search results( again)? after I export my results'. Normally to do this we would have two versions of the test, one with the new content and one with the old content, and add appropriate tags for each stage.

Unfortunately this particular step is part of another multi-step step definition, which makes that process a bit harder. Instead I've added a new step that targets the specific checkbox using the form input name (which hasn't changed).

It's a bit of a hack, but hopefully it won't be in our code for long!

### DIFF
--- a/features/buyer/tell_us_about_contract.feature
+++ b/features/buyer/tell_us_about_contract.feature
@@ -78,7 +78,7 @@ Scenario: User exports results
   And I see the 'Save a search' instruction list item status showing as 'Completed'
   When I click the 'Export your results' link
   Then I am on the 'Before you export your results' page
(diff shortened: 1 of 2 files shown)


katstevens:
Once merged, let's get another PR ready for when we deploy.

PR #758, opened 2020-05-13
attach ODT file some of the time

gidsg:
Uploads a ODT file instead of PDF 50% of the time. I realise it might not be ideal to do this in a non-deterministic way but I thought it would be quite expensive to have a whole additional flow to test this.
As I don't have much understanding of the app or domain yet would welcome any other suggestions if you have them.


katstevens:
This matches our general approach to the framework application scenario - we pick Yes/No answers at random on each run to simulate what a real user might do.

PR #767, opened 2020-06-10
Fix functional tests to reflect change in summary list markup

gidsg:
Fix tests to reflect changes in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1041

Paired with @lfdebrux


features/buyer/buyer_clarification.feature
```diff
@@ -1,4 +1,4 @@
-@buyer @buyer-clarification
+@buyer @buyer-clarification @skip-staging @skip-production
```
domoscargin, line 1: Out of curiosity, why is this skipping these stages now and not before?

lfdebrux: We need to skip them now because staging and production will be different to preview. If we make the test pass on preview, it will fail on staging.

lfdebrux: We'll take the skip tags out once the change has been deployed.

katstevens: These tests shouldn't get run on production so we probably don't need a skip tag for it (otherwise everything in this module would end up with a permanent `@skip-production`).

gidsg: @katstevens OK I made this change now in 8b2a313

PR #769, opened 2020-06-11
Fix functional tests for GOV.UK Design System flash messages in user frontend

lfdebrux:
Ticket: https://trello.com/c/sxhsW20p/88-replace-flash-messages-with-alert-component-in-briefs-frontend

Needed for alphagov/digitalmarketplace-user-frontend#265.

---

I've run these locally against local and staging.

### DIFF
--- a/features/supplier/supplier_account.feature
+++ b/features/supplier/supplier_account.feature
@@ -167,6 +167,7 @@ Scenario: Supplier user can provide and change supplier details before confirmin
   And I see 'that supplier.dunsNumber' text on the page
 
 @requires-credentials @notify
+@skip-staging
 Scenario: Supplier user can add and remove contributors
   When I click 'Contributors'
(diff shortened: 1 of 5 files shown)


features/supplier/supplier_account.feature
```diff
@@ -167,6 +167,7 @@ Scenario: Supplier user can provide and change supplier details before confirmin
   And I see 'that supplier.dunsNumber' text on the page
 
 @requires-credentials @notify
+@skip-staging
```
katstevens, line 170: Isn't this scenario on the Supplier FE?

lfdebrux: It includes the step `Then that user can not log in using their correct password` which is user frontend.

katstevens: Ah gotcha. Good spot 😸

PR #770, opened 2020-06-15
Update functional tests for Briefs FE flash messages update

domoscargin:
https://trello.com/c/sxhsW20p/88-replace-flash-messages-with-alert-component-in-briefs-frontend

Changes to the Briefs FE flash messages necessitate functional test changes.

https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/320

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -244,7 +244,7 @@ When(/^I choose a random sentence$/) do
 end
 
 Then /^I see a (success|error|notice) flash message containing #{MAYBE_VAR}$/ do |type, message|
-  flash_message = page.find(:css, ".dm-alert.dm-alert--#{type}", wait: false)
+  flash_message = page.find(:css, ".dm-alert.dm-alert--#{type}, div.flash-message-container", wait: false)
   expect(flash_message).to have_content(message)
 end
 
--- a/features/buyer/award_flow_requirements.feature
+++ b/features/buyer/award_flow_requirements.feature
@@ -13,7 +13,6 @@ Background:
   When I click the 'Let suppliers know the outcome' link for that brief
   Then I am on the 'Have you awarded a contract for %s?' page with brief 'title'
 
-
 Scenario: Award flow - Award a requirement to a winning supplier
 
   Given I choose 'Yes' radio button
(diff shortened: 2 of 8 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -244,7 +244,7 @@
 end
 
 Then /^I see a (success|error|notice) flash message containing #{MAYBE_VAR}$/ do |type, message|
-  flash_message = page.find(:css, ".dm-alert.dm-alert--#{type}", wait: false)
+  flash_message = page.find(:css, ".dm-alert.dm-alert--#{type}, div[class^='banner-']", wait: false)
```
katstevens, line 247: Nice! If we have this, do we still need the `@skip-staging` tags?

domoscargin: I don't _think_ so? I swiped the tags from #769 by @lfdebrux but then he suggested this change.

gidsg: I'm also unsure about this and exactly the order of how it fits in the workflow. If we clarify this it would be good to document it!

lfdebrux: In theory this change should mean that we don't need the tags anymore.

However when I tested this by running this branch against staging with `(. config/staging.sh ; bundle exec cucumber --strict --tags @skip-staging)` I still got failures due to the flash messages 🤔

lfdebrux: `div.flash-message-container` works better for the second selector (that's on me, typed something out without testing it before)

domoscargin: Done.

Is that `config/staging.sh` something you've created locally and does it have a specific configuration?

lfdebrux: I copied `config/common.sh` and updated the environment variables based on values in `digitalmarketplace-credentials/vars/staging.yaml`.

domoscargin: OK, this is working locally for me. Since @lfdebrux mentioned issues when removing `@skip-staging` I'm keeping those tags in and I think this should be good to merge?

lfdebrux: The tests now pass against staging, so we can remove these tags and also the existing ones in #769.

domoscargin: Done!

PR #771, opened 2020-06-15
Supplier can copy a service from a previous framework (and delete it)

katstevens:
https://trello.com/c/WBjRZd14/80-2-frameworks-repo-copyservices-list-exclude-rather-than-include-fields

We were lacking coverage of a complex part of the framework application journey, which I plan to touch on in https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1231. I thought it would be good to have some confidence that I'm not breaking anything 😸 

What we're testing:
- Supplier can see existing services from a previous framework, and copy one of them to create a new draft
- Once the draft is created, the source service 'disappears' from the list of copyable services
- Supplier can delete the draft they just created (cleaning things up nicely for our test environment!)
- Once the draft is deleted, the source service reappears in the list of copyable services

Not covered (yet...?):
- bulk copying all services from a previous framework's lot
- copying a draft service from the same framework (i.e. creating a draft from a draft)

Notes for reviewers:
- My Ruby is rusty so please feel free to suggest syntax/efficiency improvements
- Includes steps using summary tables and flash messages which we will be changing at some point, I guess/hope?
- I've tried to keep the step definitions/helpers consistent with existing methods where possible.
- `copiable`/`copyable` - both still look weird to me, not sure which one to go for 😿 
- We shouldn't need any skip tags for deploying, the following should work against existing code

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -597,6 +597,13 @@ def get_supplier_with_reusable_declaration
   JSON.parse(call_api(:get, "/suppliers/#{supplier_framework['supplierId']}").body)['suppliers']
 end
 
+def get_supplier_with_copyable_service(framework)
+  @existing_service = get_a_service("published", framework['family'])
+  puts "Service name: #{@existing_service['serviceName']}"
+  puts "Service ID: #{@existing_service['id']}"
+  JSON.parse(call_api(:get, "/suppliers/#{@existing_service['supplierId']}").body)['suppliers']
+end
+
 def remove_supplier_declaration(supplier_id, framework_slug)
   response = call_api(
     :post,
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -601,6 +601,7 @@ def get_supplier_with_copyable_service(framework)
   @existing_service = get_a_service("published", framework['family'])
   puts "Service name: #{@existing_service['serviceName']}"
   puts "Service ID: #{@existing_service['id']}"
+  puts "Lot name: #{@existing_service['lotName']}"
   JSON.parse(call_api(:get, "/suppliers/#{@existing_service['supplierId']}").body)['suppliers']
 end
 
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -404,7 +404,7 @@ def get_frameworks
   all_frameworks = JSON.parse(frameworks_from_api.body)['frameworks']
 end
 
-def get_a_service(status, framework_type = "g-cloud")
+def get_a_service(status, framework_type = "g-cloud", must_be_copyable = false)
   all_frameworks = get_frameworks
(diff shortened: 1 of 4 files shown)


features/support/api_helpers.rb
```diff
@@ -597,6 +597,13 @@ def get_supplier_with_reusable_declaration
   JSON.parse(call_api(:get, "/suppliers/#{supplier_framework['supplierId']}").body)['suppliers']
 end
 
+def get_supplier_with_copyable_service(framework)
+  @existing_service = get_a_service("published", framework['family'])
+  puts "Service name: #{@existing_service['serviceName']}"
+  puts "Service ID: #{@existing_service['id']}"
```
lfdebrux, line 603: Printing the service lot here would be helpful for debugging

features/support/api_helpers.rb
```diff
@@ -597,6 +597,13 @@ def get_supplier_with_reusable_declaration
   JSON.parse(call_api(:get, "/suppliers/#{supplier_framework['supplierId']}").body)['suppliers']
 end
 
+def get_supplier_with_copyable_service(framework)
+  @existing_service = get_a_service("published", framework['family'])
```
lfdebrux, line 601: I saw a bug running this locally where the service found was not present in the list of copyable services in the UI; I think this was because it has `copiedToFollowingFramework: true` (although I can't see the service in the list of draft services for G-Cloud 12 either). We might need some kind of way of filtering out those services.

katstevens: Hmm good shout. I'll take a look

katstevens: Added a commit - it will only look at the first page of 100 published services, but that should (hopefully!) be enough, as this scenario will clean up the copied draft once it's done.

lfdebrux:
I've stumbled on another annoying edge case (I don't know how this keeps happening to me, I've run the scenario like twice).

There is a supplier on my local machine and on preview who has multiple G-Cloud 11 services with the same name. This is because they are services that have been renamed by the [admin can edit a service scenario][1].

image

This is a problem because if the functional tests happen to pick that service to copy then it will think the copy failed because it will still see that name in the list of services to copy.

I think we could either fix this by changing the admin can edit a service scenario to change the name back, or by adding additional logic to check the href of the link.

[1]: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/efd5645be2212f557397fe81bf7e666fc47aac63/features/admin/access_a_service.feature#L7

katstevens:
> I think we could either fix this by changing the admin can edit a service scenario to change the name back, or by adding additional logic to check the href of the link.

Great spot 😸 I'll go for the href check I think - should be more robust.

katstevens:
@lfdebrux commit pushed. Hopefully this will work (glad we didn't go for pagination on this page in the end!)

PR #776, opened 2020-07-14
Update password reset/change error messages

katstevens:
https://trello.com/c/ITSfGrTU/126-1-review-error-message-content-in-user-frontend

This will need a separate PR to remove the duplicate scenarios for staging, once we're ready to deploy. One of the scenarios (logged in admin user) isn't run on staging anyway, because we don't want to send a 'real' password change email for those accounts.


katstevens:
Thanks @gidsg - let's give it a go.

PR #781, opened 2020-08-12
Account for new brief withdraw and delete process

domoscargin:
https://trello.com/c/JhIzksR3/116-2-replace-banners-with-actions-with-confirm-action-page-pattern

I'm not sure of the best way to make these work before the change and after it, or the exact process for releasing bit by bit.

### DIFF
--- a/features/buyer/requirements.feature
+++ b/features/buyer/requirements.feature
@@ -193,6 +193,65 @@ Scenario Outline: View requirement in a dashboard
     | withdrawn | Closed requirements      |
     | draft     | Unpublished requirements |
 
+
+@skip-staging
+Scenario: Delete a draft requirement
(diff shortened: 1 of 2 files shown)


lfdebrux:
In this situation it's probably best to make a copy of the original scenario with `@skip-local @skip-preview` tag, and add a `@skip-staging` tag to the new scenario. Then as the changes get released you move the tags around as required.

It's a convoluted process, but it does work. Here's an example of it in the past:

1. Add the new tests https://github.com/alphagov/digitalmarketplace-functional-tests/pull/717
2. Update tags with release to staging https://github.com/alphagov/digitalmarketplace-functional-tests/pull/720
3. Remove old tests with release to production https://github.com/alphagov/digitalmarketplace-functional-tests/pull/721

PR #783, opened 2020-08-18
Withdraw, delete confirmation pages - advance skips to staging

domoscargin:
Deletes old tests.

Removes `@skip-staging` from new tests.

See #781


lfdebrux:
If we're skipping the old tests on staging and preview we can just delete them 😁 They won't run against production, as they're not smoke or smoulder tests

PR #788, opened 2020-08-25
add functional tests for e-signature

gidsg:
This works but there are some possible improvements we could implement:

~~- [ ] set and assert lot name in contract~~ 
(better to cover this in Unit tests)
- [x] set and assert company name in contract 

~~- [ ] set and assert company address in contract~~ 
(better to cover this in Unit tests)

Co-authored-by: Paula Valenca <paula.valenca@digital.cabinet-office.gov.uk>

CC @phantas


features/step_definitions/supplier_steps.rb
```diff
@@ -100,3 +100,27 @@
...
+
+Given(/^I visit the start sign framework agreement page for that framework$/) do
+  url = "/suppliers/frameworks/#{@framework['slug']}/start-framework-agreement-signing"
+  page.visit("#{dm_frontend_domain}#{url}")
+end
+
+Then(/^I see the sign agreement confirmation page$/) do
+  confirmation_message = "You’ve signed the #{@framework['name']} Framework Agreement"
+  expect(page).to have_selector('h1', text: normalize_whitespace(confirmation_message))
+end
```
phantas, line 126:
```suggestion
end

```

gidsg: was this an accidental comment?

katstevens: Is this to address the rogue EOF?

features/supplier/supplier_signs_framework_agreement.feature
```diff
@@ -0,0 +1,23 @@
+@supplier @supplier_account @gideon_test
+Feature: Supplier can sign an agreement electronically
+
+  Background:
+    Given I have the latest live or standstill framework
+    And I have a supplier user
+    And that supplier has applied to be on that framework
+    And we accepted that suppliers application to the framework
+    And that supplier is logged in
+    And that supplier has added company details
```
katstevens: Can we rename this step to be more specific (we're only updating registered name, not any other details eg DUNS number, trading status, size...). 

Or better yet, add the `set_supplier_registered_name(@supplier["id"])` function to the `And that supplier has applied to be on that framework` step, because that's the point where this data is normally updated?

features/step_definitions/supplier_steps.rb
```diff
@@ -100,3 +100,27 @@
   @supplier = get_supplier_with_copyable_service(@framework)
   puts "supplier id: #{@supplier['id']}"
 end
+
+Given /^I have the latest live or standstill framework$/ do
+  response = call_api(:get, "/frameworks")
+  expect(response.code).to eq(200), _error(response, "Failed getting frameworks")
+  frameworks = JSON.parse(response.body)['frameworks']
+  @framework = frameworks.select { |f| f['status'] == 'live' || f['status'] == 'standstill' }.max_by { |f| f['applicationsCloseAtUTC'] }
+  expect(frameworks).not_to be_empty, _error(response, "No live frameworks found")
```
katstevens, line 109: So in our [framework application scenario step](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/step_definitions/supplier_steps.rb#L8), we have a `skip_this_scenario` method that is called if there's no `coming` or `open` frameworks. We should do the same thing here, so the scenario only gets run if the DB is in `standstill` or `live` (and more importantly, doesn't fail otherwise!).

features/support/api_helpers.rb
```diff
@@ -632,3 +632,11 @@ def set_supplier_framework_prefill_declaration(supplier_id, framework_slug, from
   )
   expect(response.code).to eq(200), response.body
 end
+
+def set_supplier_registered_name(supplier_id)
+  response = call_api(:post, "/suppliers/#{supplier_id}", payload: {
+      suppliers: { registeredName: "Awesome Company" },
+      updated_by: "functional tests",
+  })
+  expect(response.code).to eq(200), _error(response, "Failed to update supplier companies house number #{supplier_id}")
```
katstevens: We can replace this with the existing `update_supplier` function, I think?

features/supplier/supplier_signs_framework_agreement.feature
```diff
@@ -0,0 +1,23 @@
+@supplier @supplier_account @gideon_test
```
katstevens, line 1: `@gideon_test` left in here.... We should probably have a new tag along the lines of `@supplier-esignature`?

katstevens (commented):
Have left a bunch of comments - happy to chat through them if necessary?

features/supplier/supplier_signs_framework_agreement.feature
katstevens: Argh this (and the comment on `api_helpers.rb` below) were left on the wrong commit. But I think the principle is the same - we should set the supplier registered name during the 'application' step (and use the existing function to do it).

features/step_definitions/supplier_steps.rb
phantas: Yes!

features/supplier/supplier_signs_framework_agreement.feature
```diff
@@ -0,0 +1,24 @@
+@supplier @supplier_account @esignature @supplier_esignature
```
phantas, line 1: how is @esignature and @supplier_esignature different?

gidsg: I guess we could use `@esignature` e.g. if we added tests for the admin side, where `@supplier_esignature` is specific to the supplier side.

phantas: ah, thanks. I think what you have right now is good.

features/supplier/supplier_signs_framework_agreement.feature
gidsg: fixed in b2f82da

features/step_definitions/supplier_steps.rb
gidsg: fixed in 911bed0

features/support/api_helpers.rb
gidsg: good spot. Implemented in 93af6e4

features/supplier/supplier_signs_framework_agreement.feature
katstevens: Yep, looks good - no harm in having an extra granular tag

features/step_definitions/supplier_steps.rb
```diff
@@ -39,7 +39,8 @@
 
 Given /^that(?: (micro|small|medium|large))? supplier has applied to be on that framework$/ do |organisation_size|
   organisation_size ||= %w[micro small medium large].sample
-  update_supplier(@supplier["id"], 'organisationSize': organisation_size)
+  update_supplier(@supplier["id"], { 'organisationSize': organisation_size,
+                                    'registeredName': 'DM Functional Test Suppliers Ltd.' })
```
katstevens, line 43: 🏆

features/step_definitions/supplier_steps.rb
```diff
@@ -100,3 +101,27 @@
   @supplier = get_supplier_with_copyable_service(@framework)
   puts "supplier id: #{@supplier['id']}"
 end
+
+Given /^I have the latest live or standstill framework$/ do
+  response = call_api(:get, "/frameworks")
+  expect(response.code).to eq(200), _error(response, "Failed getting frameworks")
+  frameworks = JSON.parse(response.body)['frameworks']
+  @framework = frameworks.select { |f| f['status'] == 'live' || f['status'] == 'standstill' }.max_by { |f| f['applicationsCloseAtUTC'] }
```
gidsg, line 109: @katstevens Currently on Preview this will match `digital-outcomes-and-specialists-4` as it is `live`. Are we OK to move `g-cloud-12` into `standstill` before this is merged?

katstevens: The test should either work or skip accordingly - can we do a check to see if there's an available esignature framework?

gidsg: @katstevens So, I think we could do a heuristic for that like `'frameworkLiveAtUTC' >= 2020-09-28` which would match `g-cloud-12` and later frameworks.  Does that make sense?

katstevens: Yep, that should do for now 👍  We can always add a TODO, to improve if we find a better way of doing this in the future...

gidsg: OK I implemented this in 5dac6d9

features/step_definitions/supplier_steps.rb
gidsg: implemented in 2122b0d

katstevens (approved):
WFM locally with G12 in `pending`, `standstill` and `live` 🏆

PR #789, opened 2020-08-27
Be explicit about selecting radio by id

domoscargin:
https://trello.com/c/yHSxrqS9/173-3-replace-wtforms-toolkit-elements-with-govuk-frontend-components-in-award-a-dos-opportunity-journey

We select radios by ID, but with the new GOV.UK components, there's
ambiguity since the first radio's id is the same as the radios' name.

We can handle this by explicitly selecting the radio by ID, so the check
ignores the name attribute.


katstevens (approved):
Cool - `@awardflow` and `@cancel` scenarios pass for me on both master and the branch.

PR #791, opened 2020-09-04
Use Bash as Makefile shell

bjgill:
Currently we seem to use the default - sh. This causes problems with the default `local.sh` since `BASH_SOURCE` is not valid sh.

Switch to force the use of bash to ensure that `BASH_SOURCE` works.

Note that `local.sh` does have a bash shebang. However, this is useless because the shebang is not evaluated when the script is sourced.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "gemset.nix",
     "lines": null
   },
-  "generated_at": "2020-08-28T12:26:19Z",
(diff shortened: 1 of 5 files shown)


katstevens:
Is it worth taking out the shebangs from `common.sh` and `local.example.sh`?

bjgill:
Good idea.

katstevens (approved):
Good stuff 👍

PR #792, opened 2020-09-04
Don't run antivirus status test locally

bjgill:
The antivirus-api app can't meaningfully be run locally, since it relies on SNS stuff from AWS.

After this commit, the entire (un-skipped) test suite passes on my machine.


katstevens (approved):
Great spot on this 👍

PR #797, opened 2020-09-17
Add both cases to search result tests

domoscargin:
For now, we need tests using the old methods on staging and prod to pass
as well as the new tests which are on preview.


lfdebrux (approved):

lfdebrux:
Thanks @domoscargin

PR #804, opened 2020-09-28
Allow search page tests to handle new option select

domoscargin:
https://trello.com/c/VCANsToa/233-3-replace-search-filters-with-govuk-checkbox-groups-option-select

We now default to collapsing our filters, so the functional tests need to account for this behaviour.


gidsg (approved):
LGTM 🔍 👍

PR #807, opened 2020-10-05
Update 'edit company details' tests for G-Cloud 12

lfdebrux:
With G12 now live and G11 expired, our cleaned database dumps for staging and preview will have G12 as the latest live framework. The 'company details page' feature finds a supplier who is on the latest live framework, but the test still has a hardcoded reference to the name of that framework, so we need to update that whenever the latest live framework changes.


domoscargin (approved):
Ah, was just trying to figure this out!

PR #809, opened 2020-10-08
Get the framework tests working for DOS 5

bjgill:
Trello: https://trello.com/c/eLi7y2wX/423-update-functional-tests-if-necessary

We're soon going to be opening DOS 5, so it would be good to have automated tests to warn us if we inadvertently break the application journey.

These tests were last updated for G12, and did not work for DOS5. I've updated them so that they should now pass for DOS5 - see individual commits for details. I don't think I've made any changes that would break these tests for G12/3.

I've tested that these work locally. I re-ran the tests multiple times to test against a broad range of suppliers.

### DIFF
--- a/features/step_definitions/framework_application_steps.rb
+++ b/features/step_definitions/framework_application_steps.rb
@@ -160,3 +160,8 @@ Then "I see confirmation that I have removed that draft service" do
   service_name = normalize_whitespace(@existing_service['serviceName'])
   step "I see a success banner message containing '#{service_name} was removed'"
 end
+
+Then "I have submitted services for each lot" do
+  lot_count = @framework['lots'].length
+  step "I see '#{lot_count} services' text on the page"
+end
--- a/features/step_definitions/framework_application_steps.rb
+++ b/features/step_definitions/framework_application_steps.rb
@@ -112,6 +112,20 @@ Then /^I click the link to view and add services from the previous framework/ do
   step "I visit the /suppliers/frameworks/#{framework_slug}/submissions/#{lot_slug}/previous-services page"
 end
 
+# Some suppliers have multiple services with the same name. Delete all existing
+# drafts with the same name as the service we want to add to avoid later
+# trouble finding links.
+Then 'I remove existing drafts with the same name' do
+  loop do
+    draft_services_with_same_name = page.all(:xpath, "//a[contains(text(), \"#{normalize_whitespace(@existing_service['serviceName'])}\")]")
+    break if draft_services_with_same_name.empty?
+
+    draft_services_with_same_name.first.click
+    step "I click the 'Remove draft service' button"
+    step "I see 'Are you sure you want to remove this' text on the page"
+    step "I click the 'Yes, remove' button"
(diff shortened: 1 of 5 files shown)


features/step_definitions/framework_application_steps.rb
```diff
@@ -112,6 +112,20 @@
   step "I visit the /suppliers/frameworks/#{framework_slug}/submissions/#{lot_slug}/previous-services page"
 end
 
+# Some suppliers have multiple services with the same name. Delete all existing
+# drafts with the same name as the service we want to add to avoid later
+# trouble finding links.
```
lfdebrux, line 117: Hmm, I can see why you're doing this, but I wonder if we could end up missing out on opportunities to test edge cases here? Do you think there are any situations where it might be important for us to have draft services with the same name in our test databases? I'm not sure, just spitballing here.

lfdebrux (commented):
There a lot of really nice changes in here, thanks for taking the time to do this.

I just want to try running it locally against preview and staging, and then I'll approve.

lfdebrux:
I think a change in flash message content might be causing a step to fail on preview?

```
And I see '1 service will be submitted' or '1 lab will be submitted' text on the page
features/step_definitions/common_steps.rb:482
expected true
     got false
 (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/common_steps.rb:488:in `/^I (don't |)see '(.*?)'(?: or '(.*)')? text on the page/'
features/supplier/supplier_applies_to_a_framework.feature:35:in `And I see '1 service will be submitted' or '1 lab will be submitted' text on the page'
./features/step_definitions/common_steps.rb:488:in `/^I (don't |)see '(.*?)'(?: or '(.*)')? text on the page/'
features/supplier/supplier_applies_to_a_framework.feature:35:in `And I see '1 service will be submitted' or '1 lab will be submitted' text on the page'
```

User research participants was marked as complete

lfdebrux:
Works for me against staging

features/step_definitions/framework_application_steps.rb
bjgill: I agree - this is not ideal.

The problem is that if there are multiple services with the same name, I've not found a way to work out which existing service each was copied from. Do you know if there is a way of doing this?

lfdebrux: Yeah I see the problem 🤔 . It just occurred to me though that on staging it won't be so much of an issue, because we refresh it weekly. So maybe just put a `TODO` note and leave it as-is for now? It's still an improvement.

bjgill: Yes. Better imperfect tests than none.

I'll create a story to improve the tests for the framework applications and include this problem. The other thing we will want to tackle is that we don't test lots that have a maximum of only one service (which is most lots for DOS)

lfdebrux (approved):
Awesome stuff

bjgill:
Now works on preview (and I'm now set up to be able to run against preview!)

PR #810, opened 2020-10-20
Add functional test for lot limited to one service

bjgill:

In DOS, some services are limited to one service per lot (oneServiceLimit). These lots are not tested by the current tests, which only test lots with multiple services. Add a new test for these lots that will be skipped for frameworks that don't have these lots (i.e. g-cloud).

### DIFF
--- a/features/step_definitions/supplier_steps.rb
+++ b/features/step_definitions/supplier_steps.rb
@@ -96,10 +96,20 @@ Given 'I have a supplier with a reusable declaration' do
   puts "supplier id: #{@supplier['id']}"
 end
 
+Given 'I have a lot which accepts multiple services' do
+  @lot = @framework['lots'].find { |lot| lot["oneServiceLimit"] == false }
+end
+
+Given 'I have a lot limited to a single service' do
+  @lot = @framework['lots'].find { |lot| lot["oneServiceLimit"] }
+  # This sort of lot does not exist on g-cloud
+  unless @lot
+    skip_this_scenario
+  end
+end
 
 Given 'I have a supplier with a copyable service' do
-  lot_with_no_limit = @framework['lots'].select { |lot| lot["oneServiceLimit"] == false }.first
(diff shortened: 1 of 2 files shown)


features/step_definitions/supplier_steps.rb
```diff
@@ -96,10 +96,20 @@
   puts "supplier id: #{@supplier['id']}"
 end
 
+Given 'I have a lot which accepts multiple services' do
+  @lot = @framework['lots'].select { |lot| lot["oneServiceLimit"] == false }.first
```
gidsg, line 100: Could we use `find` here rather than `select` and `first`?

features/step_definitions/supplier_steps.rb
```diff
@@ -96,10 +96,20 @@
...
+end
+
+Given 'I have a lot limited to a single service' do
+  @lot = @framework['lots'].select { |lot| lot["oneServiceLimit"] == true }.first
+  # This sort of lot does not exist on g-cloud
+  if not @lot
```
gidsg, line 106:
```suggestion
  unless @lot
```

features/step_definitions/supplier_steps.rb
```diff
@@ -96,10 +96,20 @@
   puts "supplier id: #{@supplier['id']}"
 end
 
+Given 'I have a lot which accepts multiple services' do
+  @lot = @framework['lots'].select { |lot| lot["oneServiceLimit"] == false }.first
+end
+
+Given 'I have a lot limited to a single service' do
+  @lot = @framework['lots'].select { |lot| lot["oneServiceLimit"] == true }.first
```
gidsg, line 104: Can we remove the `== true` here? it may be redundant?

bjgill:
Good points all - I've applied all your suggestions and rebased against master

gidsg (approved):
👍 🥼

PR #811, opened 2020-10-20
Handle multiple services with the same name

bjgill:

A DOS supplier has multiple services with the same name. This broke a test which assumed that service names were unique for a supplier. We worked around this by deleting the services with the same name during setup (https://github.com/alphagov/digitalmarketplace-functional-tests/pull/809).

Be more discriminating about service links by getting the service ID so we can construct the href. This allows us to remove the hack and means this test now passes again for suppliers with multiple services with the same name without needing to delete them all.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "gemset.nix",
     "lines": null
   },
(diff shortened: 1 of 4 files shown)


gidsg (approved):
LGTM 👍

PR #812, opened 2020-10-21
Test submitting a copied service

bjgill:

For DOS 5, we introduced a bug that broke services that were copied from previous frameworks. This bug only showed up if you tried to complete your copied service for submission, so did not show up in our automated tests. So complete the copied service and mark it as complete during the test.

Also use `sample` so we don't use the same lot for every run of the test.


lfdebrux (approved):
This is great, I tested it against an older version of supplier frontend and it failed 😁 

Nice work 🏆

PR #814, opened 2020-10-29
Add new .app-search-summary classes

domoscargin:
https://trello.com/c/nGmcQJg6/206-1-replace-notice-panel-with-govuk-frontend-inset-text-component-in-buyer-frontend

As part of the above ticket, we replaced `search-summary` classes with `app-search-summary`, so we need to account for both in the tests.

We can remove the original classes once we've deployed.


gidsg (approved):
👍 🧷

PR #821, opened 2020-11-19
Fix functional tests on preview

lfdebrux:
We've reopened G12 on preview for UCD work, this PR fixes the functional tests so that we no longer need to hardcode any framework names or slugs in the admin company details tests.


gidsg (approved):
👍 🏑

PR #822, opened 2020-11-19
Fix functional tests for preview

lfdebrux:
More fixes for G-Cloud 12

- Simplify 'I submit a copied service' step
- Fix some corner case bugs

### DIFF
--- a/features/step_definitions/framework_application_steps.rb
+++ b/features/step_definitions/framework_application_steps.rb
@@ -56,8 +56,12 @@ Then(/^I submit a service for each lot$/) do
 end
 
 Then("I submit a copied service") do
-  answer_all_dos_lot_questions "Edit"
-  answer_all_service_questions "Add"
+  if @framework["family"] == "digital-outcomes-and-specialists"
+    answer_all_dos_lot_questions "Edit"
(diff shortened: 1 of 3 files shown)


features/step_definitions/framework_application_steps.rb
```diff
@@ -149,6 +153,7 @@
 
   new_service = get_draft_service_copied_from(@existing_service, @framework['slug'])
   @new_service_href = "/suppliers/frameworks/#{@framework['slug']}/submissions/#{new_service['lotSlug']}/#{new_service['id']}"
+  puts "Draft service URL: #{@new_service_href}"
```
bjgill, line 156: Does this need to be here?

bjgill (approved):
Looks sound. Just one question about an errant `puts`...

features/step_definitions/framework_application_steps.rb
lfdebrux: Oh, I put it in to trace down a bug, I thought it might be useful for future; they only show up if the test fails anyway.

bjgill: It's just been useful - thanks! 🙂

PR #830, opened 2020-12-17
Switch from Travis to Github Actions

bjgill:
Trello: https://trello.com/c/cnNEUjhz/2126-epic-migrate-travis-ci-usages-to-github-actions

Because Travis is now excessively slow.

Based on the starter Ruby workflow (https://github.com/actions/starter-workflows/blob/master/ci/ruby.yml)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,17 @@
+name: Test
+
+on: [pull_request]
+
+jobs:
+  test:
+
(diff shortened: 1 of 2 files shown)


domoscargin (approved):
This looks fine to me - do we need to use the `--without development` arg?

bjgill:
I don't know. It seems happy enough without, so I hope it's fine as-is? :shrug:

PR #836, opened 2021-01-07
Make functional tests work when DOS5 is live on preview

alex9smith:
Preview has DOS5 live while all other environments have DOS4 live and DOS4 in standstill. Copy the scenarios which look for a specific DOS framework name and update them for DOS5. Also remove developerPriceMin from DOS service fixture, as it's no longer a field in DOS5.

These new scenarios are only run on preview, and the existing ones are skipped on preview. This means the functional tests should pass on all environments.

Once DOS5 is live on all environments we can remove the duplicate tests as well as the DOS4 only features.

### DIFF
--- a/features/support/fixtures.rb
+++ b/features/support/fixtures.rb
@@ -9,7 +9,6 @@ module Fixtures
       lot: 'digital-specialists',
       developerLocations: %w[London Scotland],
       developerPriceMax: "200",
-      developerPriceMin: "100",
       bespokeSystemInformation: true,
       dataProtocols: true,
       helpGovernmentImproveServices: true,
--- a/features/support/fixtures.rb
+++ b/features/support/fixtures.rb
@@ -9,6 +9,7 @@ module Fixtures
       lot: 'digital-specialists',
       developerLocations: %w[London Scotland],
       developerPriceMax: "200",
+      accessibleApplicationsOutcomes: true,
       bespokeSystemInformation: true,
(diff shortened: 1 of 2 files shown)


features/support/fixtures.rb
```diff
@@ -9,7 +9,6 @@ def self.digital_specialists_service
       lot: 'digital-specialists',
       developerLocations: %w[London Scotland],
       developerPriceMax: "200",
-      developerPriceMin: "100",
```
lfdebrux, line 12: Is there any reason not to create a `digital_specialists_service_dos5` fixture as we did for DOS4?

alex9smith: Only that I think it'd be a bit messy to have a separate fixture for each framework and yet another place where schemas are defined. I'm happy to create one though if you think there are benefits

lfdebrux (approved):
Works for me! (Tested against staging)

features/support/fixtures.rb
lfdebrux: No, that makes sense. I was wondering if it would cause issues with DOS4 on staging but it seems to work fine so that's all good!

alex9smith:
Missed a change which is annoying. Retesting against preview now

alex9smith:
Ok, missed a required field out of the generic DOS fixture. New scenarios now all pass against preview and existing ones pass against staging

PR #839, opened 2021-01-11
Release alphagov/digitalmarketplace-brief-responses-frontend#255

lfdebrux:
~Reverts commit 92e1ec12cf4256cd183308fb2e54f8bd640dbc45~ Reapplies changes from #834 to staging block, as preview has since diverged further.


domoscargin (approved):
I notice most of the tests in this feature have been set to @skip-preview, so that perhaps needs to happen to these as well?

lfdebrux:
> I notice most of the tests in this feature have been set to @skip-preview, so that perhaps needs to happen to these as well?

Yeah, I think the straight revert didn't work, I couldn't get commit b860ef3 to pass against staging :( Trying a different approach.

lfdebrux:
bb9ac61 works for preview and staging

PR #840, opened 2021-01-18
Change tracking pageview step to be more specific

lfdebrux:
https://trello.com/c/ygYf9Whg/2162-functional-tests-are-failing-analytics-opt-out-tests-on-preview-and-staging

GOV.UK may send hit-only data to Google Analytics (see alphagov/frontend#2536), we need to update our tests to reflect this.

Thanks @karlbaker02 for helping triage and fix this.


lfdebrux:
Note that https://github.com/alphagov/frontend/pull/2536 is planned to be revered at some point, so this change will become unnecessary, but I think it's best to keep the test specific so we don't get caught out by any future changes.

gidsg (approved):

PR #842, opened 2021-01-22
Remove flaky test step

bjgill:
We've had this test step fail a couple of times (https://ci.marketplace.team/job/functional-tests-staging/2334/ and https://ci.marketplace.team/job/functional-tests-preview/23034/). The flash does work - I've confirmed it manually. It's just that the test is inconsistent at detecting it.

These spurious test failures mean that this test step is now providing negative value. So let's get rid of it.


alex9smith (approved):
👍

PR #844, opened 2021-02-02
Fix choose radio step for govuk-frontend

lfdebrux:
Ticket: https://trello.com/c/cMtAeSL2/764-3-add-support-for-conditionally-revealing-content-to-dmcontentgovukfronted

We need to be able to choose radio inputs for both govuk-frontend and frontend toolkit components. This commit works because both have a fieldset around the whole shebang.

This should fix the failure in https://ci.marketplace.team/job/functional-tests-preview/23057/functional_20test_20report/, I've tested it against both preview and staging (the `supplier_applies_for_an_opportunity` feature only).


gidsg (approved):
👍 🏑

PR #849, opened 2021-02-26
Make it simpler and more robust to get a supplier

bjgill:
https://trello.com/c/iDOSq28L/2180-fix-tests-that-use-i-have-a-supplier-with-step

Previously, we would get the supplier by id/name/registered name and assume it had the right data, or create it with the right data.

This caused problems because it was possible for the system to be in a state where the supplier the test got had the wrong data for the test. Also, no tests get supplier by ID.

So simplify the helper by removing the ability to get by ID.

Also make it more robust by always doing an update to ensure that the supplier we got/created has the data we expect.

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -473,41 +473,30 @@ def create_user(user_role, custom_user_data = {})
   @user
 end
 
-def get_or_create_supplier(custom_supplier_data)
-  # This method is used to search for a supplier, and if it does not exist create that supplier. It will return
-  # that supplier and also set it to the global @supplier var.
-  # The possible argument combinations for the getting of a supplier are dependent on the available end points for
-  # suppliers.
-  # Therefore this method supports:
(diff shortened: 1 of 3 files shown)


bjgill:
I've run the tests affected by this change locally and confirmed they all still pass.

features/support/api_helpers.rb
```diff
@@ -473,41 +473,30 @@ def create_user(user_role, custom_user_data = {})
...
+    supplier ||= get_supplier_by_registered_name(custom_supplier_data["registeredName"])
   end
-  if not @supplier
-    @supplier = create_supplier(custom_supplier_data)
-  end
-  @supplier
+
+  supplier ||= create_supplier
+
+  update_supplier(supplier['id'], custom_supplier_data)
```
lfdebrux, line 499: 💯

lfdebrux (approved):
This looks a lot cleaner, nice one 👍

PR #851, opened 2021-03-05
Remove unused means of getting user

bjgill:
No tests get the user using a hard-coded ID. This is good, because that would be a brittle test that was strongly coupled to the state of the database and not very reproducible. So we can remove that feature to make sure it doesn't get used in the future.


lfdebrux (approved):
👏

PR #852, opened 2021-03-09
Fix answer_all_service_questions xpath

domoscargin:
Because govukSummaryLists include hidden text in the `<a>` tag, we can't look for an exact match in text content.

But we can use `contains` to look for a partial match.


bjgill:
For reference, this was caused by https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1337

PR #853, opened 2021-03-11
Change DUNS validation scenarios to always run

lfdebrux:
Ticket: https://trello.com/c/52Xz6MJM/2184-users-are-able-to-create-supplier-accounts-with-invalid-duns-numbers

Scenarios which check our integration with the D&B DUNS service will now run even if no framework is open.

I think this is better because I think it would be good to know of any issues as soon as possible.

Scenarios which test the whole supplier application journey will still only run when a framework is open; just because it seemed overkill to me to run the whole thing every hour when it's not critical.


gidsg (approved):
🏃 👍

PR #856, opened 2021-03-15
Make the editing company details scenario more robust

lfdebrux:
This scenario tends to be flaky, and when it fails it creates an annoying scenario where the tests are in the incorrect state on the stage for the next test run to succeed.

This commit tries to work around this by making the assertions less persnickety; now the tests will only look at values after they (should) have been changed. Hopefully this will mean that if the values in the database aren't reset we won't have to log in to the admin frontend to change things manually.


gidsg (approved):
Nice one, should save us some time! Looks good to me (once rebased) 👍

PR #860, opened 2021-03-16
Make it simpler and more robust to get a user 

bjgill:
Currently, we fail the test if a supplier exists but has the wrong properties. This was a deliberate choice: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/628.

> I realize that in cases we find a property mismatch, we could actually go ahead and update the found user to match the requested properties. I've held back on doing this for now though - it has the potential to create a lot of churn on our db and allow us to introduce more race problems for parallel tests.

This has come back to bite us. We've just seen a case where a user got into the wrong state, so the tests would fail for the rest of time without manual intervention. Also, I disagree with the original author about database churn and race conditions - the only time we would do a database update is when the tests would otherwise fail, so there's not much churn. And I don't think we run our tests in parallel - and these tests couldn't be run in parallel anyway due to operating on the same users.

Thus, I think we can safely switch to new logic where we update the user if we discover that they exist but don't match what we expect. This should mean that this test becomes a bit more robust.

Note that this could cause problems if we ever used this helper method against a user in production. However, that's not something we do in any of our smoke or smoulder tests, so I think we're safe.

Follows similar logic to https://github.com/alphagov/digitalmarketplace-functional-tests/pull/849, which did the same thing for suppliers.

Also remove the ability to get user by supplier ID. It wasn't being used in any of the tests, and added complexity. It's probably a bad idea to have supplier IDs in the tests anyway for reproducibility.

### DIFF
--- a/features/support/api_helpers.rb
+++ b/features/support/api_helpers.rb
@@ -507,26 +507,24 @@ def get_or_create_supplier_with_data(name_prefix, custom_supplier_data)
   update_supplier(supplier['id'], custom_supplier_data)
 end
 
+def get_user_by_email_address(email_address)
+  response = call_api(
+    :get,
+    '/users',
+    params: { email_address: email_address }
+  )
+  JSON.parse(response.body)["users"][0]
+end
+
 def get_or_create_user(custom_user_data)
-  # This method is used to search for a user, and if it does not exist create that user. It will return that user and
-  # also sets it to the global @user var.
-  # The possible argument combinations for the getting of a user are dependent on the available end points for users.
-  # Therefore this method supports:
-  # emailAddress: email_address of the account. This can be used independently.
-  # supplierId: supplier id of the user. May result in unpredictable behaviour if there is more than one user with the
-  #     supplierId and an emailAddress is not specified
-  # Should the user not be found we will attempt a post create with the provided user data.
+  # Get the user by email address if they exist. Then ensure the user has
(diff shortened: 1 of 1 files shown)


lfdebrux:
@bjgill I agree with pretty much everything you're written above, just thought I should point out that we do run our tests in parallel when they are on Jenkins; our Makefile has a recipe [run-parallel][1] which is what gets run on Jenkins[2]

[1]: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/Makefile#L23
[2]: https://github.com/alphagov/digitalmarketplace-jenkins/blob/main/job_definitions/functional_tests.yml#L127

bjgill:
Ah. you're right that we run our tests in parallel. However, the thing we use to run tests in parallel [operates on a file level](https://github.com/grosser/parallel_tests/issues/747#issuecomment-580216980).

Thus, given that these tests are all in one file, we can be confident that they won't ever step on each other's toes.

Also, I don't think this creates any new exposure. If the tests were run in parallel prior to this PR, we'd have the problem of suppliers not being in the expected state if a conflicting test was running at the same time.

lfdebrux (approved):
This makes sense to me, it seems to work in testing, let's merge it!

bjgill:
Yes - this whole adventure shows the risk of echo-based 'testing'. Rebasing

PR #861, opened 2021-03-16
Choose a lot journey changes

domoscargin:
https://trello.com/c/r8CxfKTO/820-5-replace-choose-a-lot-page-in-create-a-service-journey

The new flow on the framework application journey includes a new page to choose a lot from a collection of radio inputs.

This adjusts the tests to account for this page.

The first commit splits out the old scenarios, so we should be able to just revert it once we've deployed to preview.


lfdebrux (approved):
We probably don't need to bother with splitting these out, as we don't have any frameworks open in any of our live stages.

PR #864, opened 2021-03-25
Fix accessible autocomplete step

lfdebrux:
Fix the step for entering text into an accessible autocomplete field.

This stopped working when we bumped accessible-autocomplete to 1.0.1 in the admin frontend[[1]]. It appears that using the `set` method doesn't work with the JavaScript for this version of the accessible autocomplete, as the typed option was not being focused.

This commit fixes it by using the `send_key` method to emulate a user typing; this works with both the newer and older version of accessible-autocomplete.

[1]: https://ci.marketplace.team/job/functional-tests-preview/23280/functional_20test_20report/

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -235,7 +235,9 @@ end
 
 When /^I enter #{MAYBE_VAR} in the '(.*)' field and click the selected autocomplete option?$/ do |value, field_name|
   field_element = page.find_field field_name
-  field_element.set value
+  field_element.set ''  # clear the field before typing
+  field_element.click
+  field_element.send_keys value
   # Find the sibling <ul>'s focused autocomplete <li> and click it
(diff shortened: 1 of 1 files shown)


lfdebrux:
I've tested this locally against preview and staging

features/step_definitions/common_steps.rb
```diff
@@ -235,7 +235,9 @@
 
 When /^I enter #{MAYBE_VAR} in the '(.*)' field and click the selected autocomplete option?$/ do |value, field_name|
   field_element = page.find_field field_name
-  field_element.set value
+  field_element.set ''  # clear the field before typing
```
domoscargin, line 238: Is there a possibility that these fields are not clear to start with?

lfdebrux: Yes, in the edit company details journey where it's an admin changing a suppliers registered address the field is populated with the existing data.

So otherwise the autocomplete will be sad because 'FranceIreland' is not a country 😂

domoscargin: _I'm_ now sad FranceIreland is not a country...

Thanks for clarifying

domoscargin (approved):
LGTM

PR #865, opened 2021-03-25
Use autocomplete step for supplier account feature

lfdebrux:
This can be merged once alphagov/digitalmarketplace-supplier-frontend#1394 is on preview.

### DIFF
--- a/features/supplier/supplier_account.feature
+++ b/features/supplier/supplier_account.feature
@@ -6,6 +6,7 @@ Background:
   And that supplier is logged in
   And I visit the /suppliers page
 
+@skip-staging
 Scenario: Supplier user can provide and change supplier details before confirming them for framework applications
   When I click 'Your company details'
   Then I am on the 'Your company details' page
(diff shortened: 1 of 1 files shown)


lfdebrux:
I've tested this PR locally and against staging.

domoscargin:
Ah, I was going to ask if this would just work by changing to using "Country" instead of "location-autocomplete" as a locator, but I see the hardcoded template has that a `span`, not a `label`, so it fails to find it.

Will merge the branch.

lfdebrux:
> Ah, I was going to ask if this would just work by changing to using "Country" instead of "location-autocomplete" as a locator, without all the @Skip stuff, but I see the old hardcoded template has "Country" inside a span, not a label, so it fails to find it.

Oh, yeah, I should have mentioned that in the commit message. Super annoying.

domoscargin (approved):
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1394 is merged

PR #867, opened 2021-03-29
Fix get_user_by_email_address

lfdebrux:
Functional tests were failing on staging after a databse refresh (when the user did not exist), see https://ci.marketplace.team/job/functional-tests-staging/2454/functional_20test_20report/.

Use `dig` method[[1]] so that if the JSON response is empty (as in the case where no users are found) then an exception will not be raised.

[1]: https://ruby-doc.org/core-3.0.0/doc/dig_methods_rdoc.html


gidsg (approved):
🧰 👍

PR #868, opened 2021-04-01
Move smoulder tests to correct directory

bjgill:
All our smoulder tests are in `features/smoulder-tests`, except one set of tests we added recently (https://github.com/alphagov/digitalmarketplace-functional-tests/pull/843). Move the new tests to the right folder to make them more readily discoverable.


gidsg (approved):
👍

PR #869, opened 2021-04-07
Skip admin data download locally

bjgill:
Trello: https://trello.com/c/XoWXM8lC/85-run-dmrunner-without-credentials

This download will never work locally. This is because this test is testing the output of a Jenkins job, which obviously doesn't happen locally.

### DIFF
--- a/features/admin/download_user_research_csvs.feature
+++ b/features/admin/download_user_research_csvs.feature
@@ -21,7 +21,7 @@ Scenario Outline: Correct users cannot view the link to download buyer user rese
     | admin-ccs-category        |
     | admin-ccs-data-controller |
 
-@file-download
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
Makes sense to me. Although if we wanted we could add a CSV file to the localstack bucket as part of [_setup_buckets](https://github.com/alphagov/digitalmarketplace-runner/blob/8a222f1080883a1825ac61d5bccb0c98a089d085/dmrunner/setup.py#L447)

bjgill:
Very true. Though I think that would reduce the value of the test. I think it's probably fine as-is - I struggle to imagine how we'd break this bit of functionality without noticing it pretty quickly in some other way.

PR #871, opened 2021-05-17
Delete feature which is unit tested

gidsg:
https://trello.com/c/asIcsC2D/584-3-convert-all-tests-that-test-admin-frontend-user-permissions-into-unit-tests

This feature is already unit tested in https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/527eca037b6ba3cbbe50e2a8899c31e2157fe746/tests/app/main/views/test_buyers.py#L20-L31 therefore we can safely remove the feature test.


bjgill (approved):
Change looks sensible and well-documented :rocket:

PR #882, opened 2021-05-20
Add command to check feature syntax and run in CI

bjgill:
We just had [a bunch of test failures](https://github.com/alphagov/digitalmarketplace-functional-tests/pull/881) because a PR was merged with a syntax error in a feature. This was not noticed before merging because the CI only linted the Ruby code.

Add a make command to parse (but not execute) all the features. This should catch future syntax errors. Run this command as part of pre-merge CI.

I've tested locally that this command succeeds for current main, and would have caught the syntax error.


gidsg (approved):
Good catch 👍 🥒

PR #884, opened 2021-05-21
remove re-testing by role

gidsg:
https://trello.com/c/asIcsC2D/584-3-convert-all-tests-that-test-admin-frontend-user-permissions-into-unit-tests

The variance in link name is already unit tested in https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/527eca037b6ba3cbbe50e2a8899c31e2157fe746/tests/app/main/views/test_index_page.py#L162L190 therefore there is no need to duplicate this in a feature test

### DIFF
--- a/features/admin/search_for_suppliers.feature
+++ b/features/admin/search_for_suppliers.feature
@@ -2,7 +2,7 @@
 Feature: Search for suppliers by registered name, DUNS number and companies house number
 
 @search-supplier-name
-Scenario Outline: Correct users search for a supplier by registered name
+Scenario Outline: Search for a supplier by registered name
   Given I am logged in as the existing <role> user
   And I have a supplier with:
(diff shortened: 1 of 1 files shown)


features/admin/search_for_suppliers.feature
```diff
@@ -17,9 +17,6 @@ Scenario Outline: Correct users search for a supplier by registered name
   Examples:
     | role                      | link-name                               |
     | admin                     | Edit supplier accounts or view services |
-    | admin-ccs-category        | Edit suppliers and services             |
```
bjgill, line 20: Does this still need to be a table?

bjgill (approved):
One question, but looks good to merge

features/admin/search_for_suppliers.feature
gidsg: resolved in 350481b

PR #888, opened 2021-06-08
update link text

gidsg:
https://trello.com/c/ohDPcudm/606-1-change-dos-page-title-user-research-studios
Reflect rename from labs to studios


bjgill (commented):
Could we describe better why we need to change this, for example by linking to the PR that caused these changes?

Also, how do you plan to release this without breaking production?

bjgill (approved):

gidsg:
@bjgill good points, I've addressed your comments now. I've added skip tags which will need to be reverted as soon as this is deployed.

PR #889, opened 2021-06-09
Add answer to `employmentStatus` question

alex9smith:
This question was added to Lots 1 & 2 of DOS in https://github.com/alphagov/digitalmarketplace-frameworks/pull/685

Add an answer in our fixtures so the automated tests can create valid briefs.

### DIFF
--- a/features/support/fixtures.rb
+++ b/features/support/fixtures.rb
@@ -188,7 +188,8 @@ module Fixtures
       title: "Tea drinker",
       workingArrangements: "Hard work",
       workplaceAddress: "London",
-      questionAndAnswerSessionDetails: "Here be Q&A details"
+      questionAndAnswerSessionDetails: "Here be Q&A details",
+      employmentStatus: "Contracted out service: the off-payroll rules do not apply"
     }
   end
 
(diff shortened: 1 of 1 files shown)


bjgill (approved):

bjgill (commented):
Is this going to break production when we merge it?

alex9smith:
Hmm actually yes it will. I'll revert the API change and have a think about how best to release

gidsg:
@alex9smith could we use `@skip-staging` & `@skip-production` tags to roll this out to the tests gradually at the appropriate time?

gidsg:
@alex9smith could we use `@skip-staging` & `@skip-production` tags to roll this out to the tests gradually at the appropriate time?

gidsg:
@alex9smith Could we use the `@skip-staging` and `@skip-production` to roll this out gradually?

alex9smith:
We _could_, but that'd mean skipping all the scenarios that involve creating, viewing or responding to a brief. I'm not sure that's a good idea since that's exactly the part of the system we're changing by adding a new question. 

I think the solution of making the question

alex9smith:
Once we have released https://github.com/alphagov/digitalmarketplace-api/pull/1146 and https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/447 this will be safe to merge again. See the description on the API PR for the explanation why.

PR #892, opened 2021-06-10
Stop testing exact framework versions

bjgill:
Some of the tests expect a particular version to be live, and so break if this changes. Update the matching to match other frameworks in the same family. This should reduce the amount of work involved in switching to a new framework.


alex9smith (approved):
Very nice!

PR #896, opened 2021-06-11
Complete new question on staging

alex9smith:
In https://github.com/alphagov/digitalmarketplace-functional-tests/pull/894 we duplicated this feature to account for the new question that's on preview but not on staging.

Once https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/447 has been released to staging, we no longer need different versions of the feature.

Remove the skip tags and update the feature so it always answers the new question.

https://trello.com/c/l9lnd1GD/14-2-add-new-ir35-question-to-the-create-an-opportunity-journey

### DIFF
--- a/features/buyer/requirements.feature
+++ b/features/buyer/requirements.feature
@@ -5,115 +5,6 @@ Feature: Create and publish a requirement
   I want to be able to create and publish a requirement
 
 
-@skip-preview @skip-local
-Scenario: Create individual specialist requirement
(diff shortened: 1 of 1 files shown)


alex9smith:
I'm not planning on merging this until the briefs FE has been released to staging, but it'd be easier to have it reviewed and approved so it's all ready to go when we need it so the pipeline's not blocked.

bjgill (approved):

PR #899, opened 2021-06-24
use find rather than select 1st

gidsg:
https://trello.com/c/uIVs4QP6/47-2-timebox-speed-up-i-am-logged-in-as-the-buyer-of-a-closed-brief-with-responses-steps

Minor refactor found during the course of the above story. ﻿


bjgill (approved):
Neat (in both senses of the word :slightly_smiling_face: )

PR #900, opened 2021-07-01
Update to Capybara v3

bjgill:
Start the process of moving us towards the present day - the current version is 3.35.3.

I think 3.0.3 should be safe - no obvious breaking changes that affect us in v3. And PhantomJS was still supported as of 3.0.3.

One deprecation that we'll need to attend to:

```
DEPRECATED: Capybara::Helpers::normalize_whitespace is deprecated, please update your driver
```


bjgill:
Draft whilst I test

bjgill:
I think I've found the fix: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/903

bjgill:
A lot closer - all the buyer tests now pass, except for those to do with downloading files.

bjgill:
Though that is something that is quite different between selenium and phantomjs. So maybe we should switch off phantomjs first?

bjgill:
We did the upgrade in https://github.com/alphagov/digitalmarketplace-functional-tests/pull/912

PR #901, opened 2021-07-01
Add support for running headless with Selenium

bjgill:
https://trello.com/c/OcwZuCo5/48-2-timebox-improve-reliability-of-selecting-an-autocomplete-option

I have a hypothesis that using Selenium might fix some of the performance/reliability issues we're currently seeing. To test this, I want to run an experiment where we try running the tests with Selenium and see what happens.

First, though, we need to add an option to run Firefox in headless mode. This is because it would be utterly pointless to run the user interface on Jenkins.


alex9smith (approved):
👻

PR #903, opened 2021-07-02
Stop using Capybara to normalise whitespace

bjgill:
Capybara 3 has deprecated this method, so we need to start normalising whitespace ourselves.


gidsg (approved):
📏 👍

PR #906, opened 2021-07-22
Make cookie settings tests work in ChromeDriver

gidsg:

This tackles a couple of issues which prevented the cookie-settings tests working in ChromeDriver:
* Reset sessions did not clear cookies unless we are on the domain under test
* the `network_traffic` method is not available in ChromeDriver, therefore we use the ChromeDriver performance logs API to provide equivalent functionality

(see commit extended messages for full details)

## Testing

If you have `ChromeDriver` installed:

```
BROWSER=true CHROME=true DM_ENVIRONMENT=preview make ARGS='--tags @cookie-settings' run
```
should pass.

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -3,6 +3,7 @@ require 'securerandom'
 require 'uri'
 
 Given /^I (?:re-?)?visit the homepage$/ do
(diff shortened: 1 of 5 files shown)


alex9smith (approved):
This looks sensible and runs locally as suggested for me!

PR #907, opened 2021-07-23
Make @file-download tests work in ChromeDriver

gidsg:

Previously there were a couple of issues preventing the file download tests working in ChromeDriver:

* By default Chrome shows a user prompt before downloading a file, this PR sets a preference to disable that.
* The `response_headers` method is not available in ChromeDriver. Therefore similarly to #906 this uses the Chrome performance logs as a replacement to derive this data.

## Testing

If you have `chromedriver` installed.

Running:
```
BROWSER=true CHROME=true DM_ENVIRONMENT=preview make ARGS='--tags @file-download' run
```
should pass

### DIFF
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -544,16 +544,21 @@ Then(/^I should get an? (download|inline) file(?: with file.?name ending(?: in)?
     target_window = current_window
   end
   if is_chrome
-    disposition_parts = inline_http_requests.map { |r| r.dig('content-disposition') }.compact[0].split(";")
+    requests = inline_http_requests
+    disposition_parts = requests.map { |r| r.dig('content-disposition') }.compact[0].split(";")
   else
     within_window(target_window) do
       disposition_parts = page.response_headers['Content-Disposition'].split(";")
     end
-    expect(disposition_parts[0]).to eq(case download_inline when "download" then "attachment" else download_inline end)
-    if ending
-      expect(disposition_parts[1]).to match("^\s*filename=.*#{Regexp.escape(ending)}['\"]?\s*$")
-    end
-    if content_type
+  end
+  expect(disposition_parts[0]).to eq(case download_inline when "download" then "attachment" else download_inline end)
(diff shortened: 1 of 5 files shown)


features/step_definitions/common_steps.rb
```diff
@@ -543,9 +544,12 @@
...
   end
-
-  within_window(target_window) do
-    disposition_parts = page.response_headers['Content-Disposition'].split(";")
+  if is_chrome
+    disposition_parts = inline_http_requests.map { |r| r.dig('content-disposition') }.compact[0].split(";")
+  else
+    within_window(target_window) do
+      disposition_parts = page.response_headers['Content-Disposition'].split(";")
+    end
```
bjgill, line 552: Why are we ending the `within_window` here?

features/support/helpers.rb
```diff
@@ -240,3 +240,29 @@ def get_summary_list_rows_by_preceding_heading(heading)
...
+  if is_chrome
+    # Chrome does not support network_traffic, instead we can extract this from the performance logs
+    logs = page.driver.browser.manage.logs.get(:performance)
+    # Store messages in a structure which is easier to work with
+    messages_array = logs.each_with_object([]) do |entry, messages|
+      message = JSON.parse(entry.message)
+      timestamp = entry.timestamp
+      messages << message
+      message.store(:timestamp, timestamp)
+    end
```
bjgill, line 258: I don't quite understand what's going on here - could we have a quick chat?

bjgill (commented):
Looks sensible - just one bit I'm struggling to understand.

features/step_definitions/common_steps.rb
gidsg: fixed in 7e1157b4134b0f73e55a4924d56638b43029e830

features/support/helpers.rb
gidsg: @bjgill improved in ad4f126

bjgill (approved):
:tada:

PR #909, opened 2021-07-26
Tweaks to fix autocomplete input in ChromeDriver

gidsg:
Couple of tweaks to make the country autocomplete work more reliably in ChromeDriver.

* Change the method of clearing the input
* Slow down the text input to allow the JavaScript to fire

## Testing
If you have ChromeDriver installed:
```
BROWSER=true CHROME=true make ARGS='--tags @company-details-edit' run
```
should pass.

Also:
```
make ARGS='--tags @company-details-edit' run
```
should still pass (i.e. this is backwards compatible with PhantomJS)


daniele-occhipinti (approved):
I've run the test against dmp-compained and it passed. Tested two things in one. Double yeah! :-)

PR #910, opened 2021-07-26
Fix Live Reload test in ChromeDriver

gidsg:

`\n` was not interpreted as a return keypress in ChromeDriver, e.g.
download (2)

This fixes it by sending a return keypress event.

## Testing

If you have ChromeDriver installed:
```
BROWSER=true CHROME=true make ARGS='--tags @live-reload' run
```
should pass

Should also be backwards compatible with PhantomJS i.e.
```
make ARGS='--tags @live-reload' run
```
should pass


gidsg:
...the tests pass on this branch https://ci.marketplace.team/job/functional-tests-preview/23655/

PR #912, opened 2021-07-28
Upgrade Capybara and switch to ChromeDriver 

gidsg:
https://trello.com/c/MmkirLrD/76-3-switch-from-phantomjs-to-chrome

PhantomJS is end-of-life, we want to switch to a supported driver and ChromeDriver seems the most feature rich, fast and stable.

This PR takes a 'big bang' approach because the latest version of Capybara makes ChromeDriver more stable but this breaks backwards compatibility with PhantomJS:

* Upgrade to latest version of Capybara
* Remove all conditional logic and makes Chrome/ChromeDriver sole supported engine
* Assorted fixes to support these changes (see commit messages for details)

As this is 'big bang' it can be reverted in its entirety if we want to go back to the old behaviour.

I have run several builds on Jenkins to check for flakiness and it seems stable: https://ci.marketplace.team/job/Functional%20Tests%20-%20ChromeDriver/


alex9smith:
This looks sensible to me and I agree making one big change is the right thing to do here. 

In cddabac the commit message says about upgrading to Capybara `3.0.3` but actually changes to `3.7.2`, and in 2ad6401 you upgrade again to `3.35.3`. Is that what you were intending to do?

gidsg:
> In cddabac the commit message says about upgrading to Capybara 3.0.3 but actually changes to 3.7.2, and in 2ad6401 you upgrade again to 3.35.3. Is that what you were intending to do?

Ah, good spot @alex9smith I think this reflects the evolution of this PR. @bjgill went to the minimum version which had a fix he wanted but I just went to the latest version, which is the approach we decided on now with this 'big bang' there are not concerns about retaining backwards compatibility. I will see if the commit history can be cleaned up at all.

alex9smith:
Makes sense - I hadn't realised it'd started out as a minimum possible change. Happy to approve once you've sorted that!

gidsg:
@alex9smith ...OK I've dropped that first commit now and re-worded the other one, so hopefully the commit history makes more sense now.

alex9smith (approved):
LGTM!

PR #918, opened 2022-01-04
Get Design System 3 page count

bjgill:
The previous xpath worked for the old toolkit, but not for the new Design System 3. We've now switched over all search pages to Design System 3, so don't need to be back-compatible.

The previous behaviour was concealing a bug because the xpath never found anything, so it assumed that there was only one page of results.

I've tested this locally against Preview, and it fails as expected (due to the bug currently exposed on Preview).

### DIFF
--- a/features/support/catalogue_helpers.rb
+++ b/features/support/catalogue_helpers.rb
@@ -5,13 +5,9 @@ module CatalogueHelpers
 
   def self.get_page_count(page)
     begin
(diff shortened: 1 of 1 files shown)


tim-s-ccs (approved):

bjgill:
Hurrah. The database resync has hidden the bug again, and this test now passes. Merging.

PR #923, opened 2022-03-25
Bump Ruby to 2.7.0

bjgill:
This matches the system version of Ruby on our Jenkins server. Start using this version so that we can use the system Ruby on Jenkins instead of relying on rbenv, which doesn't work for us as of the latest Jenkins version.

Seems to work for me locally, but the functional tests are rather broken on my system, so I won't know for sure until we merge and run them on Jenkins.


tim-s-ccs:
I have couple of trick when I need to re-trigger a build.

If I'm not too bothered about the commit message I will do:
- `git commit --amend` and the change the message
- `git push -f` which will then force push the branch up to GitHub

If I don't want to affect the commit message I will do:
- `git log` to view the latest commit and get its number
- `git reset --hard HEAD~1` which will undo that commit
- `git cherry-pick $commit-number` which then gets the branch back to the state I want it
- `git push -f` which will then force push the branch up to GitHub

tim-s-ccs:
Yes, they've been doing maintenance today which has affected GitHub actions on Pull Requests (but not actions on the push I noticed). The maintenance has now finished but there still seems to be some lingering issues.

